PR #449, opened 2015-02-10
fixed data changed behaviour when not reloading in itembox

danrot:
Fixed data changed behaviour, so that the event is also thrown when list is not reloaded.

Added also another method for empty data check, so that the default can also be overriden by the concrete implementation.


wachterjohannes:
@danrot ACK i will test it in sulu!

PR #452, opened 2015-02-17
[FEATURE] added validation for time to input types

michaelzangerle:
__Tasks:__
- [x] done
- [x] got feedback

__Informations:__

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| Tests pass?   | none
| Fixed tickets | none
| Doc           | none


turbo-ele:
ack

dist/husky.js
```diff
@@ -26702,23 +26702,27 @@ define('type/husky-input',[
...
                     return regex.test(value);
                 },
-                url:  function(value) {
+                url: function(value) {
                     var regex = /^([\da-z\.-]+)\.([a-z\.]{2,6})([\/\w\.-\?\$\#])*\/?/;
                     return regex.test(value);
                 },
-                color:  function(value) {
+                time: function(value) {
+                    var regex = /^([01]?[0-9]|2[0-3]):[0-5][0-9]$/;
```
wachterjohannes: @michaelzangerle why do you have done this exactly? this is only for german format ... in the content we would like to configure it this is possible with the validation library ... but this overrules them all ...

michaelzangerle: @wachterjohannes this has nothing to do with a german format it's simply a 24 hour time format which should also work internationally. According to Oliver we do not make am/pm formatted time ... If you are asking me why it is in here and not in husky validation then you have to ask the person who implemented this component!

wachterjohannes: we have actually said that we use the according "normal" format  of the country and for my opinion in english is time with PM/AM and thats what the validation component does actually it changes the behaviour of validation according selected language!

if we want to change the format for a specific language (here english) we should try to override the [Culture-Information](https://github.com/jquery/globalize/tree/v0.1.1#defining-culture-information) for this language! That would be done in the Validation library!

michaelzangerle: well thats one opinion vs another. I personally have no problem with an am/pm format but thats what we needed for our task. Discuss it with Oliver or change it to whatever you would like to have but for our task it has to be a 24 hour format...

PR #459, opened 2015-02-24
Added time validation

wachterjohannes:
__Tasks:__


husky_components/input/input-type.js
```diff
@@ -54,29 +50,60 @@ define([
...
+                default: function() {
+                    return this.$el.find('input').val();
+                },
+                time: function() {
+                    var value = typeGetter.default.call(this),
+                        dateTime = Globalize.parseDate(value, this.options.timeFormat);
+
+                    return Globalize.format(dateTime, 'HH:mm:ss');
+                },
+                date: function() {
```
danrot: Why is the date only in the getter, but not in the setter?

wachterjohannes: thats because the component is not very good you have to get the value from inside the component and there is no public api for that ... set is no problem you can put it on the data of the element ...

PR #461, opened 2015-02-26
Enable empty time field

wachterjohannes:
__Tasks:__

__Informations:__

| Q             | A
| ------------- | ---
| Tests pass?   | yes
| Fixed tickets | none
| BC Breaks     | none
| Doc           | none


danrot:
I don't really know what this is about, tried to leave an empty value in the time field, but it is still failing at the validation.

PR #462, opened 2015-02-27
Bugfix/style

marcelmoosbrugger:
- Added a medium loader to datagrid, which can be shown and hidden via events
- Added a selected-counter for datagrid
- removed search from navigation


husky_components/datagrid/decorators/table-view.js
```diff
@@ -286,10 +292,29 @@ define(function() {
...
+                this.sandbox.dom.removeClass(this.datagrid.$find('.selected-elements'), constants.counterIndentClass);
+            }
+        },
+
+        /**
          * Destroys the view
          */
         destroy: function() {
             this.sandbox.stop(this.sandbox.dom.find('*', this.$el));
+            this.removeIndentSelectedCounter();
```
danrot: Is this necessary? The entire element gets destroyed afterwards anyway, doesn't it?

marcelmoosbrugger: the view gets destroyed. but i added the select-counter to the main-class as it's needed in other views as well. in the case where the table-view takes up the whole width a had to give the counter a little indent on the right as it looked horrible without. Therefore i have to remove the indent after the table-view gets destroyed. These things are the purpose of the destory-methods of views

PR #466, opened 2015-03-24
Added table-view render without root elements

wachterjohannes:
Belongs to https://github.com/sulu-io/sulu/pull/923 (CRM extraction)


danrot:
Why was this change necessary?

wachterjohannes:
because we want to display a part of the tree and this was no possible because the mapping always uses the  parent of a category

PR #467, opened 2015-04-10
Coffee icon only show when image not exists

alexander-schranz:


husky_components/datagrid/decorators/thumbnail-view.js
```diff
@@ -233,6 +233,14 @@ define(function() {
...
                 this.selectItem(id);
             }.bind(this));
+
+            this.sandbox.dom.on(this.$thumbnails[id].find('img'), 'error', function() {
+                this.$thumbnails[id].find('img').remove();
+            }.bind(this));
+
+            this.sandbox.dom.on(this.$thumbnails[id].find('img'), 'load', function() {
+                this.$thumbnails[id].find('.fa-coffee').remove();
+            }.bind(this));
```
wachterjohannes: and in the table view? what is shown there?

alexander-schranz: added coffee icon fallback to tableview

PR #470, opened 2015-04-13
Selected item in column view

wachterjohannes:
fixes https://github.com/sulu-io/sulu/issues/929


wachterjohannes:
i have to check this solution with @berndhep

wachterjohannes:
@danrot i have discussed this with bernd! this is ok for him!

PR #472, opened 2015-04-14
changed tab data to match content navigation

danrot:


dist/husky.js
```diff
@@ -2074,7 +2074,7 @@ define("requirejs", function(){});
 // the stack via arguments.caller.callee and Firefox dies if
 // you try to trace through "use strict" call chains. (#13335)
 // Support: Firefox 18+
-//
+//"use strict";
```
dantleech: Remove commented code

danrot: Usually I am with you, but this isn't even our code ;-)

PR #475, opened 2015-04-20
Fixed focus and enter save for data-grid table-view

wachterjohannes:
__tasks:__

- [ ] submit changes to the documentation

__informations:__

| q                | a
| ---------------- | ---
| Fixed tickets    | fixes sulu-io/sulu#1018
| BC breaks        | none
| Documentation PR | none


danrot:
@3rfan What do you think about that change? Should we add information like a field is currently edited in the data attributes?

husky_components/datagrid/decorators/table-view.js
```diff
@@ -1060,15 +1060,21 @@ define(function() {
...
                 attribute = this.sandbox.dom.data(this.sandbox.dom.parents($inputs[0], 'td'), 'attribute');
             }
+
             $cell = this.table.rows[recordId].cells[attribute].$el;
-            this.sandbox.dom.show(this.sandbox.dom.find('.' + constants.inputWrapperClass, $cell));
-            this.sandbox.dom.focus(this.sandbox.dom.find('.' + constants.editableInputClass, $cell));
-            this.sandbox.dom.select(this.sandbox.dom.find('.' + constants.editableInputClass, $cell));
+            $inputField = this.sandbox.dom.find('.' + constants.editableInputClass, $cell);
+            $inputWrapper = this.sandbox.dom.find('.' + constants.inputWrapperClass, $cell);
+            this.sandbox.dom.data($inputField, 'edit', 'on');
```
danrot: @chirimoya @3rfan Would you put information like this into data attributes?

3rfan: @danrot u know how to do it, baby ;)

danrot: @wachterjohannes Wouldn't it be a good idea to just remove the enter handler? Would solve the issue with having multiple requests in a better way.

wachterjohannes: but then the enter doesn´t work then ... do you realy think thats the solution?

danrot: I think it will continue to work, because currently you have two requests, one for the enter, and one for the blur. If enter triggers the blur it will still work. At least that's what I remember from our conversations.

danrot: Just tested, it works without the listener for the enter button. The only problem I see is the `isFocusoutHandlerEnabled` option. You could register the enter handler only if this option is not enabled.

husky_components/datagrid/decorators/table-view.js
```diff
@@ -1092,8 +1099,10 @@ define(function() {
         editableInputFocusoutHandler: function(event) {
             if (!!this.isFocusoutHandlerEnabled) {
                 this.sandbox.dom.stopPropagation(event);
-                var recordId = this.sandbox.dom.data(this.sandbox.dom.parents(event.currentTarget, '.' + constants.rowClass), 'id');
-                this.editRow(recordId);
+
+                var e = jQuery.Event('keypress');
```
wachterjohannes: @danrot @3rfan what do you think about this change?

3rfan: Imo a better approach would be to create a states list for all the inputs and use the input state to check if you need to save the entry.

danrot: @3rfan Actually Apart from the fact that a global variable is used here I would think this approach is better than handling another state manually... The only advantage I see in your solution is that we could save all remaining changes once instead immediately after every change, which we don't do currently... /cc @chirimoya

PR #477, opened 2015-04-28
fixed scroll bug

3rfan:


husky_components/toolbar/main.js
```diff
@@ -647,7 +647,7 @@ define(function() {
         setButtonWidth = function(listItem, parent) {
             var maxwidth = 0, i, length;
             if (parent.type === 'select') {
-                for(i = -1, length = parent.items.length; ++i < length;) {
+                for (i = -1, length = parent.items.length; ++i < length;) {
```
3rfan: remove the last semicolon. this results in a syntax error.

3rfan:
Else Ack

husky_components/toolbar/main.js
chirimoya: `;` is required

PR #480, opened 2015-05-05
Fixed event timing bug for datagrid

wachterjohannes:
Event should be thrown later. because if someone routes by this event every time an exception is thrown! because dom already is cleaned


wachterjohannes:
@danrot ping

danrot:
@wachterjohannes Where can this bug be reproduced?

wachterjohannes:
@danrot i have fixed this everywhere in sulu! but this should be fixed for future usages

PR #483, opened 2015-05-14
Overlay: Tabs component changed structure of data

wachterjohannes:
it was forgotten to change it in overlay.

@danrot @chirimoya eventually we should release it as HOTFIX ? because in sulu the version tab of media is missing because of that!


danrot:
Yes, we have to release that as a bugfix IMO.

PR #484, opened 2015-05-19
Fixed toolbar border if first element is not displayed

chirimoya:


3rfan:
Acked and ty :dancer:

PR #492, opened 2015-06-19
Fixed ckeditor dialog in ie

3rfan:
Added Ckeditor IE dialog css and and fixed the styling for IE:

__informations:__

| q                | a
| ---------------- | ---
| Fixed tickets    | fixed https://github.com/sulu-io/sulu/issues/1223


danrot:
@3rfan Why are there changes in the `dialog.css` file now? I guess these changes are causing some errors in the other browsers... (some fields are not appearing anymore in the demo).

3rfan:
@danrot The changes should only affect the submit button inside the the footer of the dialog. Could you describe the browser errors more precisely?

bower_components/ckeditor/skins/husky/dialog.css
```diff
@@ -243,6 +243,15 @@ div.cke_disabled .cke_dialog_ui_labeled_content div * {
...
 
+.cke_dialog_ui_hbox_first {
+    display: none;
+}
+
+.cke_dialog_ui_hbox_last {
+    padding: 5px;
+    text-align: center;
+}
+
```
danrot: @3rfan These are the changes I am talking about. They are not in the IE file, so why do they need to be here? If you check e.g. the overlay for inserting a link now in a browser like chrome you will recognize that there are just 2 form fields instead of 4.

3rfan:
@danrot fixed

PR #503, opened 2015-07-15
style fixes datagrid and tabs

marcelmoosbrugger:
__tasks:__

- [x] make the whole cell containing the edit icon clickable and therefore black in font color.
- [x] remove the load effect datagrid
- [x] in tabs: adjust the marker when items get removed or added

__informations:__

| q                | a
| ---------------- | ---
| Fixed tickets    | none
| BC breaks        | none
| Documentation PR | none


demos/datagrid/script.js
```diff
@@ -53,9 +52,9 @@ require(['lib/husky'], function(Husky) {
                     matchings: 'http://husky.lo:7878/admin/api/datagrid/fields',
                     selectedCounter: true,
                     selectedCounterText: 'selected elements',
-                    clickCallback: function(item, data) {
+                    /*clickCallback: function(item, data) {
                         app.logger.log('Husky.Ui.DataGrid item click: ' + item, data);
-                    },
+                    },*/
```
danrot: Remove this code if it is not required. But I would leave it there, or did you have a reason to remove it?

husky_components/datagrid/main.js
```diff
@@ -1414,11 +1414,15 @@
 
             /**
              * Updates the selected-elements counter
-             * @param {String|number} number - the number of selected items
              */
-            updateSelectedCounter: function(number) {
+            updateSelectedCounter: function() {
```
danrot: Shouldn't this change affect already existing code?

marcelmoosbrugger: no its only internal. Its more convenient if i read the length of an array within the method instead of passing the number around

demos/datagrid/script.js
marcelmoosbrugger: fixed

PR #546, opened 2015-09-15
Fixed bug in itembox and added events to datagrid

wachterjohannes:
Main PR: https://github.com/sulu-io/sulu/pull/1347


danrot:
What was this bug causing?

husky_components/datagrid/main.js
```diff
@@ -495,6 +495,24 @@
             },
 
             /**
+             * deselects an item with a given id
+             * @event husky.datagrid.deselect.item
+             * @param {String|Number} The id of the item to deselect
+             */
+            DESELECT_ITEM = function() {
+                return this.createEventName('deselect.item');
```
danrot: Don't we have the verb usually last?

husky_components/datagrid/main.js
```diff
@@ -495,6 +495,24 @@
...
+            },
+
+            /**
+             * toggle an item with a given id
+             * @event husky.datagrid.toggle.item
+             * @param {String|Number} The id of the item to toggle
+             */
+            TOGGLE_ITEM = function() {
+                return this.createEventName('toggle.item');
```
danrot: Same as above, would prefer `item.toggle`, if it is not a big deal.

husky_components/datagrid/main.js
wachterjohannes: hmm all other events in the datagrid has the same style (for my opinion also wrong) - like `husky.datagrid.select.item`. and i think we should`t change all event names ...

husky_components/datagrid/main.js
wachterjohannes: same as above

wachterjohannes:
a request is done with the ids array was empty ... then the loader was not removed and the before items are in the list ... all of this can happen if you deselect all items in overlay!

husky_components/datagrid/main.js
danrot: Ok, then it is just different for all other components. But I agree, changing that would be a too big BC break.

husky_extensions/itembox.js
```diff
@@ -327,7 +327,7 @@ define(function() {
                 // reset items visible when new content is loaded
                 this.viewAll = false;
 
-                if (!!data) {
+                if (!!data && data.length > 0) {
```
wachterjohannes: created new issue with media-selection (is not a array its an object ...)

danrot: I hope the media-selection is working anyway? And can you link the issue here, for reference?

wachterjohannes: yeah media-selection is a object so it has no length (:

PR #553, opened 2015-09-28
Feature/datagrid set data

janmassive:


demos/datagrid/script.js
```diff
@@ -13,6 +13,61 @@ require(['lib/husky'], function(Husky) {
...
+                        "name": "title"
+                    },
+                    {
+                        "translation": "videomanagement.chapters.start",
+                        "name": "start"
+                    },
+                    {
+                        "translation": "videomanagement.chapters.end",
+                        "name": "end"
+                    }
```
danrot: I guess you should use a more general example than `videmanagement`(because of the translations). I would go with `id`, `title` and `description`.

demos/datagrid/script.js
```diff
@@ -498,6 +545,18 @@ require(['lib/husky'], function(Husky) {
                     app.logger.log('Husky.Ui.DataGrid item action: ' + item, data);
                 }
             }
+        },
+        ,
```
danrot: I guess there should only be one comma.

PR #555, opened 2015-09-29
[FEATURE] Select preselect

turbo-ele:


dist/husky.js
```diff
@@ -38518,10 +38520,12 @@ define('__component__$select@husky',[], function() {
             // Used as a fallback to revert to the last committed data
             this.mergedData = null;
 
-            // when preselected elements is not set via options look in data-attribute
-            if (!this.options.preSelectedElements || this.options.preSelectedElements.length === 0) {
+            // if data value is written,
+            if (!this.options.preSelectedElements ||
+                this.options.preSelectedElements.length === 0 ||
+                this.options.preferPreselectionFromData
```
wachterjohannes: why do you need that option?

turbo-ele:
add default option

husky_components/select/main.js
```diff
@@ -74,7 +75,8 @@ define([], function() {
             resultKey: '',
             showToolTip: false,
             translations: translations,
-            isNative: false
+            isNative: false,
+            default: null
```
wachterjohannes: default is a keyword use `defaultValue` instead ...

wachterjohannes:
else ACK

PR #556, opened 2015-09-30
[BUGFIX] added check if slice exists before cropping text

turbo-ele:


wachterjohannes:
ACK i will try it in husky

PR #560, opened 2015-10-05
Feature/itemgrid

janmassive:


husky_extensions/itemgrid.js
```diff
@@ -0,0 +1,311 @@
...
+            // click on the config button
+            this.sandbox.dom.on(this.$deleteButton, 'click', function() {
+                this.sandbox.emit(this.DELETE_BUTTON_CLICKED());
+            }.bind(this));
+        },
+
+        addItem = function(newItem) {
+            var data = [];
+
+            if (App.util.typeOf(this.getData()) === 'array' && App.util.typeOf(this.getData()[0]) === 'object') {
```
danrot: You should not be using the `App` variable here, please use `this.sandbox` instead.

husky_extensions/itemgrid.js
```diff
@@ -0,0 +1,311 @@
...
+                data = this.getData();
+            }
```
danrot: I don't really understand the logic here at all... I would write it something like this:

```javascript
var data = this.getData();

if (this.sandbox.util.typeOf(data) === ...) {
    data = [];
}
```

This way the `getData` method would only be called once, which is good, because I would assume that this method could also have a serious performance impact, depending on the type of list it is used with.

janmassive: Ok, but don't you mean !== in the truth assignment?

danrot: Of course :smile:

PR #562, opened 2015-10-08
Added url-input component

wachterjohannes:
Main PR https://github.com/sulu-io/sulu/pull/1670


husky_components/url-input/main.js
```diff
@@ -0,0 +1,232 @@
...
+         * Set data to dom-data and generate URL.
+         *
+         * @param {{scheme, specificPart}} data
+         */
+        setData: function(data) {
+            var newData = this.sandbox.util.extend(true, {}, this.getData(), data);
+
+            newData.url = null;
+            if (!!newData.scheme && !!newData.specificPart) {
+                newData.url = this.getUrl(newData);
```
danrot: I would not write this information on this object, since at can be recreated out of the other two properties at any time. Duplicating this information just raises the probability that an error occurs.

wachterjohannes: i will change that.

danrot:
You still have the url properties on these data objects... wouldn't it be cleaner to just don't duplicate this information, and let the validation type return the combined value?

wachterjohannes:
@danrot fixed

PR #564, opened 2015-10-08
[FEATURE] Added option for hiding overlay header; improved select to define mul…

turbo-ele:
…tiple styles


dist/husky.js
```diff
@@ -38562,6 +38562,26 @@ define('__component__$select@husky',[], function() {
...
+        /**
+         * Process styles of options and apply to select.
+         *
+         * @param {Object} $button
+         */
+        processStyles = function($button) {
+            if (!!this.options.style) {
+                var styles = this.options.style.split(' ');
+                this.sandbox.util.foreach(styles, function(style) {
+                    if (style === 'small') {
```
turbo-ele: in_array

husky_components/select/main.js
```diff
@@ -257,6 +257,26 @@ define([], function() {
         },
```
turbo-ele: options - docu

PR #568, opened 2015-10-14
Fixed to avoid missing util functions

wachterjohannes:


wachterjohannes:
@danrot what did you think?

husky_extensions/util.js
```diff
@@ -9,7 +9,7 @@
  */
 
 /**
- * @deprecated use util-service instead
+ * Provides the util function into a sandbox.
  */
 define(['services/husky/util'], function(Util) {
```
danrot: Would write `util` here, if you start saying it is not a constructor :wink:

husky_extensions/util.js
```diff
@@ -19,53 +19,9 @@ define(['services/husky/util'], function(Util) {
...
-            app.core.util.removeFromArray = Util.removeFromArray;
-
-            app.core.util.capitalizeFirstLetter = Util.capitalizeFirstLetter;
-
-            app.core.util.arrayMap = Util.arrayMap;
-
-            app.core.util.object = Util.object;
+            for (var key in Util) {
+                app.core.util[key] = Util[key];
+            }
```
danrot: It doesn't work something like:

```javascript
app.core.util = Util;
```

wachterjohannes: no because the app.core.util is an object which is already initialized from aura and has a few functions from there. so i have to extend this object.

wachterjohannes:
@danrot fixed

husky_extensions/util.js
```diff
@@ -19,53 +19,9 @@ define(['services/husky/util'], function(Util) {
...
-
-            app.core.util.removeFromArray = Util.removeFromArray;
-
-            app.core.util.capitalizeFirstLetter = Util.capitalizeFirstLetter;
-
-            app.core.util.arrayMap = Util.arrayMap;
-
-            app.core.util.object = Util.object;
+            for (var key in util) {
+                app.core.util[key] = Util[key];
```
danrot: Shouldn't it be here also `util`?

wachterjohannes: hmm no exception in demo ... strange ...

wachterjohannes:
@danrot ping (:

PR #569, opened 2015-10-19
Added callback action for itemgrid

janmassive:


danrot:
Looks like you are missing a build here.

PR #577, opened 2015-11-12
Added set-options to column-navigation

wachterjohannes:


husky_components/column-navigation/main.js
```diff
@@ -930,6 +941,22 @@ define([], function() {
             }
         },
 
+        unBindDOMEvents: function() {
```
danrot: `unbind`

PR #580, opened 2015-11-13
added events to data navigation for selecting items

danrot:
Reverts the change from #576, and added events to data-navigation, so that it is possible to properly select an item via events (instead of emulating a click, which has loads of side effects).


bower_components/aura/lib/ext/mediator.js
```diff
@@ -92,7 +91,6 @@ define('aura/ext/mediator', function () {
        * @param {Object} payload    Payload emitted
        */
       app.sandbox.emit = function () {
-        if (this.stopped) { return; }
```
danrot: These lines prevented me from emitting an event in a callback. I have looked up the aura.js repo, and these lines are gone there also, and `this.stopped` doesn't seem to be used in this file elsewhere, so I guess that should be fine.

wachterjohannes: hmm i´m not sure if we should remove that ... this would enable emitting events on "lost" components ... i would take a look why you try to emit an event on a stopped component ...

wachterjohannes:
@danrot i would take a look on the emit stopped ... rest is ok!

bower_components/aura/lib/ext/mediator.js
danrot: As said, this line is also missing in the original library, so it was probably removed there as well: https://github.com/aurajs/aura/blob/master/lib/ext/mediator.js#L93

It doesn't look like the component is lost, otherwise it would not work, I guess.

wachterjohannes: i know we have some trouble sometimes where these lines was the "main" actor. There we try to emit an event in a callback of a stopped component ... so if we remove them we enable these behavior where stopped and (maybe) lost components can emit events.

danrot: So this lines were added by us?

wachterjohannes: no they where here when we fork the repository. But we have a lot of trouble with lost components in the past. therefor i feel not well when we remove these lines ...

danrot: I can say that it prevents this line from working: https://github.com/sulu-io/sulu/pull/1761/files#diff-54853adb238ed957cca321cfa254b7c1R101

I guess that's something that should be possible, since this sandbox and its corresponding component are still existing afterwards. Do you have any references where this might make trouble? Then we could check on these as well (and maybe find another solution).

wachterjohannes: if stopped is true then the sandbox was stopped before ... maybe someone closes the data-navigation before? i think we have "removed" all lost components but i dont want to get into troubles with that because we dont know if there will be some lost components in the future ....

PR #581, opened 2015-11-18
Fixed input component change event

wachterjohannes:
Main PR: https://github.com/sulu-io/sulu/pull/1800


danrot:
Why is this necessary?

wachterjohannes:
because for the preview we have to fire this update event from the container of a component because only on this the preview will listen

husky_components/input/main.js
```diff
@@ -200,6 +200,13 @@ define([], function() {
          * Binds Dom-related events
          */
         bindDomEvents: function() {
+            this.sandbox.dom.on(this.$el, 'change', function(event) {
+                event.stopPropagation();
+                event.preventDefault();
+
+                this.$el.trigger('change');
```
danrot: Referring to your other comment: What you are basically doing here is to catch the event emitted on `this.$el`, and triggering again, so how can this improve the situation?

wachterjohannes: you have overseen the filter in the next line (: i catch the event from the input not from the $el!

PR #584, opened 2015-11-30
[ENHANCEMENT] Implemented reset functionality of toolbar-button

turbo-ele:
__tasks:__

- [ ] kind of awesome
- [ ] awesome
- [ ] super awesome
- [ ] awesome awesome
- [x] maybe awesome

__informations:__

| q                | a
| ---------------- | ---
| Fixed tickets    | none
| BC breaks        | none
| Documentation PR | none


husky_components/toolbar/main.js
```diff
@@ -1081,10 +1134,15 @@ define(function() {
...
 
+                // set default title and icon
+                item.defaultTitle = item.title;
+                item.defaultIcon = item.icon;
+
                 // save to items array
                 this.items[item.id] = item;
                 this.items[item.id].initialized = dfd.promise();
 
+
```
wachterjohannes: remove this empty line

husky_components/toolbar/main.js
```diff
@@ -663,7 +680,43 @@ define(function() {
...
+            // reset title
+            item.title = this.sandbox.translate(item.defaultTitle);
+            this.sandbox.dom.html(listItems.eq(1), item.title);
+            this.items[itemId].title = item.title;
+
+            // remove marked class from dropdown-item
+            if (!!item.dropdownItems) {
+                var dropdown = this.sandbox.dom.find(selectors.dropdownMenu + ' li', listElement);
+                this.sandbox.dom.removeClass(dropdown, 'marked');
```
wachterjohannes: you can use the selector here without the find in the line before

husky_components/toolbar/main.js
```diff
@@ -663,7 +680,43 @@ define(function() {
...
+            }
+
+            if (this.options.responsive === true) {
+                updateOverflow.call(this);
+            }
+
+            this.sandbox.emit(BUTTON_CHANGED.call(this));
```
wachterjohannes: was this event already there before?

turbo-ele: yes, i didn't event that :)

don't like that?

wachterjohannes: hmm its ok but how was it used before? are you sure that you dont trigger any "unexpected" stuff?

husky_components/toolbar/main.js
```diff
@@ -362,6 +353,35 @@ define(function() {
...
 
         /**
+         * Changes a button text and icon.
+         *
+         * @param {String} button
+         * @param {String|Number} id
+         * @param {Bool} executeCallback
+         */
+        changeButton = function(button, id, executeCallback) {
+            if (!!this.items[button]) {
```
wachterjohannes: early return here:

```
if (!!this.items[button]) {
    return;
}

...
```

husky_components/toolbar/main.js
turbo-ele: when calling 'husky.toolbar.item.change' this has been triggered before, yes. 

since i'm still calling the same event, it's more consistent to still trigger it.

what i was also thinking about is, was to specify an event 

`husky.toolbar.item.reset` instead of using the change item. what do you think about that?

wachterjohannes: i would trigger that functionality with a own event i think ... `husky.toolbar.item.reset` the it would be clear what does happen

turbo-ele: allright

husky_components/toolbar/main.js
```diff
@@ -362,6 +365,56 @@ define(function() {
...
+                    resetMainListItem.call(this, button.$el);
+                    return;
+                }
+
+                // update icon
+                var index = getItemIndexById.call(this, itemId, button);
+                changeMainListItem.call(this, button.$el, button.dropdownItems[index]);
+                this.sandbox.emit(ITEM_MARK.call(this), button.dropdownItems[index].id);
+                if (executeCallback === true || !!button.dropdownItems[index].callback) {
+                    if (typeof button.dropdownItems[index].callback === 'function') {
```
wachterjohannes: combine this ifs

husky_components/toolbar/main.js
```diff
@@ -362,6 +365,56 @@ define(function() {
...
+                // if id is null - unset icon and title when no item selected
+                if (itemId === null) {
+                    resetMainListItem.call(this, button.$el);
+                    return;
+                }
+
+                // update icon
+                var index = getItemIndexById.call(this, itemId, button);
+                changeMainListItem.call(this, button.$el, button.dropdownItems[index]);
+                this.sandbox.emit(ITEM_MARK.call(this), button.dropdownItems[index].id);
```
wachterjohannes: do you really want to mark this item when you change it?

husky_components/toolbar/main.js
```diff
@@ -362,6 +365,56 @@ define(function() {
...
+                }
+            }.bind(this));
+        },
+
+        /**
+         * Resets button to its original state.
+         *
+         * @param {String} buttonId
+         */
+        resetButton = function(buttonId) {
```
wachterjohannes: reset all buttons would not be interesting?

husky_components/toolbar/main.js
turbo-ele: this was here before, i don't want to change that

husky_components/toolbar/main.js
turbo-ele: not now

PR #586, opened 2015-12-03
Enhanced table-view

wachterjohannes:
Main PR: https://github.com/sulu-io/sulu/pull/1851


husky_components/datagrid/main.js
```diff
@@ -1084,6 +1086,37 @@
             },
 
             /**
+             * Parses data and returns normalized embedded items.
+             *
+             * @param data
+             * @param parent
```
danrot: Please add types.

husky_components/datagrid/main.js
```diff
@@ -1084,6 +1086,37 @@
...
+             */
+            parseEmbedded: function(data, parent) {
```
danrot: What purpose does this have? Does it create a tree, like what is done until now for the categories? Maybe call it `parseTree` instead, so that it is more clear?

wachterjohannes: but it does more than this (currently at least normalizing id property)

PR #590, opened 2015-12-16
[ENHANCEMENT] Added event when datagrid calls url that returns an error

turbo-ele:


danrot:
What do you need that for?

turbo-ele:
to be able to react if datagrid throws an error

PR #592, opened 2015-12-18
Trigger error event when datagrid cannot load an url

turbo-ele:


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # CHANGELOG for husky
 
+* Unreleased
```
wachterjohannes: we use dev-master for that

PR #594, opened 2016-01-18
Changed style of warning label

danrot:
__informations:__

| q                | a
| ---------------- | ---
| Fixed tickets    | none
| Related PRs      | https://github.com/sulu-io/sulu/pull/1233
| BC breaks        | none
| Documentation PR | none


dist/husky.css
```diff
@@ -10562,22 +10562,29 @@ a.btn-black, a.btn-highlight {
...
-  top: 0px;
-  left: 0px;
+  top: 0;
+  left: 0;
   color: #fff;
-  line-height: 70px;
   padding-left: 55px;
   font-size: 12px;
-  z-index: 15000;
+  z-index: 200;
```
danrot: This was necessary to fix this Bug:
image

I think a z-index value that high is not required anymore, since the new labels push the content away, instead of overlapping it. Or do you know any use cases where this might make problems? /cc @wachterjohannes @chirimoya 

PS: I think we should adapt [something like this](https://css-tricks.com/handling-z-index/), adjusting z-index values is a PITA currently.

wachterjohannes: you totally right the z-indexes are widely distributed in the project. whenever you change something it could make problems anywhere. but i have no idea where it could make problems (: we will see!

PR #597, opened 2016-02-01
Fixed css bugs and added features to datagrid

wachterjohannes:
Main PR https://github.com/sulu-io/sulu/pull/1927


husky_components/datagrid/decorators/table-view.js
```diff
@@ -442,7 +444,7 @@ define(function() {
             if (!!this.options.selectItem && !!this.options.selectItem.type) {
                 var $cell = this.sandbox.dom.createElement(templates.headerCell);
                 this.sandbox.dom.addClass($cell, constants.checkboxCellClass);
-                if (this.options.selectItem.type === selectItems.CHECKBOX) {
+                if (this.options.selectItem.type === selectItems.CHECKBOX && this.options.disabledItems.length === 0) {
```
danrot: What is rendered if there are some `disabledItems`?

danrot: Or the better question is: Shouldn't there be a possibility to select all non-disabled items?

husky_components/select/main.js
```diff
@@ -1100,14 +1101,18 @@ define([], function() {
...
-                this.sandbox.emit(EVENT_SELECTED_ITEM.call(this), key);
+                this.sandbox.emit(EVENT_SELECTED_ITEM.call(this), key, selectedValue);
             }
         },
 
         // triggers select callback or emits event
         triggerPreSelect: function(key) {
+            if (!this.options.emitPreSelect) {
+                return;
+            }
```
danrot: Why do we need to configure if this event is emitted? Couldn't we simply not listen to the event? If there is no good reason for that, I would omit it, since it makes stuff unnecessarily complicated.

danrot:
@wachterjohannes We should also create better PRs in Husky, a simple link to the main PR is not enough. What exact css bugs are fixed here? That's probably not described anywhere, or is it?

CHANGELOG.md
```diff
@@ -1,10 +1,14 @@
...
 
 * dev-develop
+<<<<<<< HEAD
     * BUGFIX      #616 Fixed label tick in firefox
     * ENHANCEMENT #615 Introduced no-img-icon as function to use record dependent icons
     * BUGFIX      #613 Fixed search-icon to prevent routing
     * FEATURE     #610 Added save parameter for datagrid
+=======
+    * FEATURE     #597 Fixed css bugs and added features to datagrid
+>>>>>>> c820cfb... added disabled items in table view
```
danrot: Please fix this conflict.

husky_components/datagrid/decorators/table-view.js
wachterjohannes: i have introduced this feature but i don't need it in the master PR i will extract it into an own PR which can be discussed separately.

husky_components/select/main.js
wachterjohannes: i will introduce a preselect callback here to avoid an additional flag.

PR #600, opened 2016-02-03
Bugfix/globalize datetime

### LINKED ISSUE #496: Bug in globalize extension on formating datetime
https://github.com/massiveart/husky/blob/develop/husky_extensions/globalize.js#L74

Flag returnDateOnly is used wrong. Line should look like this:
"if (returnDateOnly !== true) {"

in the current implementation, the function return the time only when returnDateOnly is true. But exactly the opposite case is correct.

rs2487:
Fixes #496


rs2487:
Fixes #496

PR #605, opened 2016-02-19
Added loader to indicate loading suggetions

wachterjohannes:
__informations:__

| q                | a
| ---------------- | ---
| Fixed tickets    | none
| Related PRs      | none
| BC breaks        | none
| Documentation PR | none


husky_components/auto-complete/main.js
```diff
@@ -71,10 +71,13 @@ define([], function() {
...
             '<div class="husky-auto-complete <%= dropdownSizeClass %>">',
-                '<div class="front">',
-                    '<a class="fa-<%= autoCompleteIcon %>"></a>',
-                '</div>',
-                '<div class="input"></div>',
+            '    <div class="front">',
+            '       <a class="fa-<%= autoCompleteIcon %>"></a>',
+            '    </div>',
+            '    <div class="input"></div>',
+            '    <div class="back">',
```
danrot: Why class `back` doesn't sound like a good description to me.

wachterjohannes: this is already introduced into the input component of husky

danrot: Forget it for now, the other class is also called front, and it was probably like that previously.

PR #611, opened 2016-03-03
Added style for label-description

wachterjohannes:
Style for input-description

__informations:__

| q                | a
| ---------------- | ---
| Fixed tickets    | none
| Related PRs      | none
| BC breaks        | none
| Documentation PR | none

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
     * FEATURE     #610 Added save parameter for datagrid
     * BUGFIX      #612 Fixed link style in white-boxes
     * ENHANCEMENT #614 Added evaluate tab-conditions
+    * ENHANCEMENT #611 Added style for label-description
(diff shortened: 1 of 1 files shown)


scss/modules/form.scss
```diff
@@ -40,7 +40,7 @@ input[type="text"] {
 input
 .input-description {
     margin-top: -5px;
-    font-size: 10px;
+    font-size: $smallFontSize;
```
danrot: Doesn't this change the font size of a lot of different fields? Is that intended?

wachterjohannes: i have discussed this with bernd and he said that it should be 12px instead of 10px. this was not used in sulu before this and custom-url feature.

danrot:
@wachterjohannes What happened to this PR? Looks like it can be closed without merging.

wachterjohannes:
it looks like i have done the same changes in an other PR. so i will close this.

PR #613, opened 2016-03-07
Fixed search-icon to prevent routing

wachterjohannes:
Main PR: https://github.com/sulu-io/sulu/pull/2091


husky_components/search/main.js
```diff
@@ -22,8 +21,8 @@ define([], function() {
 
     var templates = {
             skeleton: [
-                '<a class="fa-search fa-flip-horizontal search-icon" href="#"></a>',
-                '<a class="fa-times-circle remove-icon" href="#"></a>',
+                '<span class="fa-search fa-flip-horizontal search-icon"></span>',
+                '<span class="fa-times-circle remove-icon"></span>',
```
danrot: Would you really make that spans? `<a>` tags would indicate that this is something clickable (making the application more accessible).

wachterjohannes: the problem is that you always have todo some weird stuff with `preventDefault` and `return false` which makes the event listener more complicated and hard to maintain. changing that to a `span` makes it much more useable. i have also added the css to show a pointer on this container.

wachterjohannes: after a short offline discussion i will try to use a button here.

danrot: It might make it easier to maintain, but it will definitely be less usable. It's impossible to use screenreaders or similar tools like that. And shouldn't a `onclick="return false;"` in the html be enough?

wachterjohannes: thats also weird ... but i will test it! like i said for me its not an anchor like it is used here because we don't change the URL or do any routing. a button would be the semantic right at this point

wachterjohannes:
@danrot i have changed it to buttons which works also in ie9 (:

PR #614, opened 2016-03-08
Added evaluate tab-conditions

wachterjohannes:
Main PR: https://github.com/sulu-io/sulu/pull/2099


husky_components/tabs/main.js
```diff
@@ -290,8 +315,51 @@ define(function() {
...
+         *
+         * @returns {boolean}
+         */
+        evaluate: function(displayConditions, conditionData) {
+            for (var i = 0, length = displayConditions.length; i < length; i++) {
+                var item = this.sandbox.util.extend(true, {}, {
+                        property: null,
+                        operator: null,
+                        value: null
+                }, displayConditions[i]);
```
danrot: wrong indentation

husky_components/tabs/main.js
```diff
@@ -196,6 +206,22 @@ define(function() {
             setMarker.call(this);
         },
 
+        update = function(conditionData) {
+            this.conditionData = conditionData;
```
danrot: Is it really necessary to introduce this state here?

husky_components/tabs/main.js
```diff
@@ -290,8 +315,51 @@ define(function() {
...
+                        break;
+                    case 'neq':
+                        if (conditionData[item.property] === item.value) {
+                            return false;
+                        }
+                        break;
+                    default:
+                        this.sandbox.logger.error('operator "' + item.operator + '" is not implemented.');
+
+                        return false
```
danrot: `;` missing

husky_components/tabs/main.js
```diff
@@ -50,6 +50,7 @@ define(function() {
             forceReload: false,
             callback: null,
             forceSelect: true,
+            conditionData: {},
```
danrot: Moving the discussion from https://github.com/sulu-io/sulu/pull/2099#discussion_r55395972 to here. What about renaming it to `formData`? And if I am right with https://github.com/massiveart/husky/pull/614/files#r56297942 we could also theoretical rename it do `data`, although having `this.data` and `data` is also anything else than optimal :confused:

wachterjohannes: hmm also formData is not correct here. because every kind of data could be passed here. i think conditionData is not "nice" but it describes it quite good. they are data to check the conditions.

wachterjohannes:
@danrot everything fixed except the "conditionData" question

husky_components/tabs/main.js
danrot: I don't really think so, sounds more like the condition themselves are passed. Usually it is about forms, that's why I came up with this. What about `passedData` or `testData` (because the described conditions are tested against this data).

danrot: And maybe we should not forget that we could call it just `data` now, which would still be better than `conditionData` IMO.

wachterjohannes: but that is not the "data" which is used to render the component - at least in all other components we use the term "data" like this.

PR #615, opened 2016-03-09
Introduced no-img-icon as function to use record dependent icons

wachterjohannes:
Main PR: https://github.com/sulu-io/sulu/pull/2103/files


danrot:
changelog missing.

PR #618, opened 2016-03-23
Fixed update of record with a different id property

wachterjohannes:
If you try to add and update (without reload) a record in a datagrid where you have a different ID property.


wachterjohannes:
This PR fixes the problem with disappearing custom-urls. i will create a PR where this is in

PR #619, opened 2016-03-24
Introduced info text for dropdown

wachterjohannes:
This PR added info texts for dropdown items. Additionally the stopPropagation is removed to use this event outside of the dropdown (sulu for copy to clipboard https://github.com/sulu-io/sulu/pull/2155).


husky_components/dropdown/main.js
```diff
@@ -33,6 +33,8 @@ define([], function() {
             shadow: true,  // if box-shadow should be shown
             toggleClassOn: null, // container to set is-active class
             valueName: 'name', // name of text property
+            infoName: 'info', // name of info property (used to display hover info)
+            clickedName: 'clickedInfo', // name of info property (used to display clicked info)
```
danrot: Why name? What is this actually doing? Are these the translations?

husky_components/dropdown/main.js
```diff
@@ -151,36 +154,61 @@ define([], function() {
...
+            if (!item[this.options.clickedName]) {
+                return this.hideDropDown();
+            }
+
+            $item.addClass('highlight-animation');
+            $item.find('.info, .info-clicked').addClass('clicked');
+
+            _.delay(function() {
+                $item.removeClass('highlight-animation');
+                $item.find('.info, .info-clicked').removeClass('clicked');
```
danrot: I would save the result from the call a few lines earlier instead of calling `find` again.

husky_components/dropdown/main.js
```diff
@@ -151,36 +154,61 @@ define([], function() {
...
+                return this.hideDropDown();
+            }.bind(this), 1000);
```
danrot: Instead of `1000` we could use an option here?

husky_components/dropdown/main.js
wachterjohannes: they are property names

husky_components/dropdown/main.js
```diff
@@ -291,7 +337,25 @@ define([], function() {
...
+            // ignore it if the target is children of trigger but not children of item
+            if (!!event
+                && ((
+                    $(event.target).hasClass('husky-dropdown-trigger')
+                    || $(event.target).parents().hasClass('husky-dropdown-trigger')
+                )
+                && !(
+                    $(event.target).hasClass('husky-dropdown-item')
+                    || $(event.target).parents().hasClass('husky-dropdown-item')
+                )
```
danrot: Can't the previous two `&&` be connected to one? Maybe some filter that checks if the element has one class and doesn't have another one? Maybe also the two next statements can somehow be included like this.

husky_components/dropdown/main.js
```diff
@@ -291,7 +337,25 @@ define([], function() {
...
+                || $(event.target).hasClass('highlight-animation')
+                || $(event.target).parents().hasClass('highlight-animation'))
+            ) {
+                this.sandbox.dom.one(this.sandbox.dom.window, 'click', this.hideDropDown.bind(this));
```
danrot: Looks like the comment above is wrong, it does not simply ignore this case. (Although I don't really understand what it does at all).

husky_services/util.js
```diff
@@ -290,6 +290,23 @@ define(function() {
         return string.charAt(0).toUpperCase() + string.slice(1);
     };
 
+    /**
+     * Returns human readable byte string.
+     * 
+     * @param {int} bytes
+     * @returns {String}
+     */
+    Util.prototype.formatBytes = function(bytes){
```
danrot: Nice :smiley:

husky_components/dropdown/main.js
danrot: And what to these properties contain?

wachterjohannes: a info for a hover and the info when you click on the item

husky_components/dropdown/main.js
```diff
@@ -291,7 +341,19 @@ define([], function() {
         },
 
         // hide dropDown
-        hideDropDown: function() {
+        hideDropDown: function(event) {
+            // do not hide if the target is children of trigger but not children of item or the element was highlighted
+            // reapply the hide listener
+            if (!!event
+                && ($(event.target).is('.husky-dropdown-trigger:not(.husky-dropdown-item), .highlight-animation')
+                    || $(event.target).parents().is('.husky-dropdown-trigger:not(.husky-dropdown-item), .highlight-animation')
```
danrot: Didn't you put the result of `$(event.target)` into its own variable.

husky_components/dropdown/main.js
```diff
@@ -151,36 +154,64 @@ define([], function() {
...
+                return this.hideDropDown();
+            }
+
+            $infos = $item.find('.info, .info-clicked');
+            $item.addClass('highlight-animation');
+            $infos.addClass('clicked');
+
+            _.delay(function() {
+                $item.removeClass('highlight-animation');
+                $infos.find('.info, .info-clicked').removeClass('clicked');
```
danrot: Shouldn't the `find('.info, .info-clicked')` part be removed here?

wachterjohannes: fixed

PR #622, opened 2016-03-29
Added navigation divider for sections with no title

wachterjohannes:
This PR adds a divider for navigation sections without a title. Additionally it adds the support to define default-messages. For example if a message in FR is not there it would take the default messages (EN for example).

This code will be used in https://github.com/sulu-io/sulu/pull/2196


husky_extensions/globalize.js
```diff
@@ -57,7 +57,7 @@
                         return key;
                     }
 
-                    return !!translation ? translation : key;
+                    return !!translation || translation === '' ? translation : key;
```
danrot: Some brackets would probably make more clear what's happening here. And what does it actually do?

wachterjohannes: it return the value if it is not null or an empty string.

PR #625, opened 2016-04-04
Improved regex for urls

wachterjohannes:
Improved regex for URLs.

fixes sulu/sulu#2174


husky_components/input/input-type.js
```diff
@@ -9,8 +9,8 @@
  */
 
 define([
-    'type/default'
-], function(Default) {
+    'type/default','services/husky/url-validator'
```
danrot: Don't we usually put these into separate lines?

husky_components/input/input-type.js
```diff
@@ -9,8 +9,8 @@
...
+], function(Default, UrlValidator) {
```
danrot: Didn't we agree to start services with lower letters, since they are not constructors?

wachterjohannes: your right (:

wachterjohannes:
@danrot fixed

PR #636, opened 2016-04-18
Added possibility to define size of label

danrot:
This PR introduces a new option to define the size of a label. Currently there is an option called `size` taking strings (such as `'big'` or `'small'`), which are passed as modifier classes to the created element.

@chirimoya @wachterjohannes Do you like that solution? Any better ideas?


scss/modules/label.scss
```diff
@@ -1,35 +1,42 @@
...
     position: relative;
     top: 0;
     left: 0;
     color: $white;
-    padding-left: 55px;
     font-size: $smallFontSize;
     z-index: 200;
 
+    &.big {
+        height: 70px;
```
danrot: Thought about adding this one as default (meaning declaring it a level higher). Didn't really like it, I thought the label must always have a size class. Feels more consistent.

wachterjohannes:
in other cases we use the term `skin` why not using this also here? i also dont like it to much but we shouldnt fragment the "meanings" of options to much in my opinion.

danrot:
Because it has a huge drawback: The number of skins (or types, which is basically the labels name for skins) would grow rapidly. If we integrate the size also in these types, we would already have: `error`, `error_small`, `warning`, `warning_small`, `success`, `success_small`, `success_icon`, `success_icon_small`. I'd prefer to also remove these types, so that you can combine these properties as you wish, without too much maintenance overhead from our side.

wachterjohannes:
but there is the type for warning, error and success. in the overlay we have the same:

````
 * @params {String} [options.skin] set an overlay skin to manipulate overlay's appearance. Possible skins: '', 'dropzone'
 * @params {String} [options.slides[].type] The type of the overlay ('normal', 'error' or 'warning')
```

skin can also be wide. this is quite the same as you want to achieve. maybe not the best solution but should be changed in both components if we want to change the behavior.

danrot:
The types in the label are a little bit different... They are just a predefined set of options, which will be merged into the options array later. So they are already not the same as in the overlays. And I really don't want to build it that way just because it is like that somewhere else.

scss/modules/label.scss
danrot: Another solution would be to do something like `&-big` and use `@extend .husky-label-warning` and only adapt the height there. Then we would only adapt the height there. But it might be harder to understand, and makes it maybe a little bit harder to change the size of the label dynamically, in case that will ever be a requirement.

wachterjohannes:
the types in the overlay works exactly the same (: therefor it is only a naming issue in my opinion. i would change the name of `size` into `skin` or let the size be `70px` for example.

danrot:
That's the point: It is not only a naming issue. I don't want to specify so many types, because it will end up in something like `error_option1_option2_option3`. Actually I thought it would also make sense to remove the types here at all. Then we would have a simple option for everything configurable, and Sulu could combine these options to something useful in the [labels component](https://github.com/sulu/sulu/blob/1c8861ce26c4f407b23a60ba475b0dd93fc5d53f/src/Sulu/Bundle/AdminBundle/Resources/public/js/components/labels/main.js). But that would a too big refactoring right now.

danrot:
In that case I could also live with passing pixels as size, although having classes would also have advantages here (think about a redesign, we would only have to adapt the css value for this class, instead of hoping that everybody used the same abstraction so that we can change it there).

husky_components/label/main.js
```diff
@@ -47,7 +48,8 @@ define(function() {
         vanishDuration: 250,
         showDuration: 250,
         closeCallback: null,
-        insertMethod: 'append'
+        insertMethod: 'append',
+        size: 'big'
```
chirimoya: use `this.options.labelClass` instead for now

chirimoya:
ACK

PR #637, opened 2016-04-22
[BUGFIX] Load correct globalize.culture file

trickreich:
Is needed for specific locales like 'de_CH'.
The globalize culture files are named like 'globalize.culture.de-CH'


wachterjohannes:
@trickreich ACK but a changelog is missing

PR #643, opened 2016-04-26
Added ckeditor plugin registry

wachterjohannes:
This PR adds the ability to register ckeditor plugins.

**Example:**

In a aura-extension:

```
sandbox.ckeditor.addPlugin(
    'internalLink',
    {
        init: function(editor) {
...
        }
    }
);
sandbox.ckeditor.addToolbarButton('links', 'InternalLink');
```


husky_extensions/ckeditor-extension.js
```diff
@@ -23,7 +23,7 @@
         }
     });
 
-    define(['ckeditor', 'jqueryAdapter'], function() {
+    define(['underscore', 'ckeditor', 'jqueryAdapter'], function(_) {
```
danrot: Why do we need `'ckeditor'` and `'jqueryAdapter'` here, although it is not passed to the function?

husky_extensions/ckeditor-extension.js
```diff
@@ -33,28 +33,155 @@
...
+        ToolbarBuilder.prototype.add = function(name, items) {
+            if (!this.toolbar[name]) {
+                this.toolbar[name] = [];
+            }
+
+            this.toolbar[name] = this.toolbar[name].concat(items);
+        };
+
+        /**
+         * Add toolbar items.
```
danrot: Wrong comment

husky_extensions/ckeditor-extension.js
```diff
@@ -33,28 +33,155 @@
...
+         *
+         * @param {String} name
+         * @param {String[]} items
```
danrot: Is that `[]` specified in YUIDoc?

husky_extensions/ckeditor-extension.js
```diff
@@ -33,28 +33,155 @@
...
+                this.toolbar[name] = [];
+            }
+
+            this.toolbar[name] = _.difference(this.toolbar[name], items);
+        };
+
+        /**
+         * Returns toolbar.
+         *
+         * @returns {{name, items}[]}
```
danrot: Is that kind of YUIDoc a real thing? Shouldn't it just be `Object`? You could define in the comment how the content looks in more detail.

husky_extensions/ckeditor-extension.js
```diff
@@ -33,28 +33,155 @@
...
+
+                    addToolbarButton: function(toolbarName, button, icon) {
+                        if (!toolbar[toolbarName]) {
+                            toolbar[toolbarName] = [];
+                        }
+
+                        toolbar[toolbarName].push(button);
+
+                        if (!!icon) {
+                            icons.push(icon);
```
danrot: `icons` is an object, right? Then it probably doesn't have a `push` method.

husky_extensions/ckeditor-extension.js
```diff
@@ -33,28 +33,155 @@
...
+                        }
+                    },
+
+                    getToolbar: function() {
+                        return JSON.parse(JSON.stringify(toolbar));
```
danrot: Use `this.sandbox.util.deepCopy` instead.

husky_extensions/ckeditor-extension.js
wachterjohannes: we need it to be sure that it is loaded by require. this scripts will add some global vars and constructor's we need.

husky_extensions/ckeditor-extension.js
wachterjohannes: we could use Object but then everything can be passed this is more specific. it says it is an object with the property name and items.

husky_extensions/ckeditor-extension.js
wachterjohannes: thats right (:

husky_extensions/ckeditor-extension.js
wachterjohannes: i have it from the google style guide https://google.github.io/styleguide/javascriptguide.xml#JsTypes

wachterjohannes:
@danrot fixed

PR #650, opened 2016-05-17
Added hide tabs for only one single tab

wachterjohannes:
This PR adds the functionality to hide tabs which has only one visible.

This changes are used by the https://github.com/sulu/sulu/pull/2367


danrot:
I guess you are talking about tabs here, and not tags 😉 

And how did that work before? If the form for adding a page was opened before the tabs were also not visible.

wachterjohannes:
you are totally right (:

danrot:
Then it would be nice if you could also update the commit, in case somebody is reading that later 😉 

And you completely ignored the second part of my comment 😛 Why is this change necessary at all? It also seemed to be working before, without the need to pass the number of visible tabs and do something in the other component?

wachterjohannes:
@danrot done :)

wachterjohannes:
@danrot i will check this tomorrow.

wachterjohannes:
closed in favor of https://github.com/massiveart/husky/pull/652

PR #659, opened 2016-06-08
Added own implementation for paste-from-word plugin

wachterjohannes:
This PR add a very simple but complete implementation for the paste-from-word plugin. We want to avoid any formation which was introduced in word.


husky_components/ckeditor/plugins/paste-from-word/main.js
```diff
@@ -0,0 +1,83 @@
...
+                                    {
+                                        type: 'cancel'
+                                    },
+                                    {
+                                        type: 'ok',
+                                        text: 'OK'
+                                    }
+                                ],
+                                okCallback: function() {
+                                    var text = this.$el.find('textarea').val();
```
danrot: Is the fact that we use a `textarea` removing all the stuff we don't want? Or where else is that happening?

chirimoya:
@wachterjohannes ping

wachterjohannes:
@chirimoya @danrot this was a test to make the PasteFromWord a bit more "strict" i will rebase it and maybe you can test it and if you think its good we can merge.

husky_components/ckeditor/plugins/paste-from-word/main.js
wachterjohannes: yes we use the testarea (: which works for quite good for me.

danrot:
Do I have to have a special word file for testing that? Or what exactly does the old solution have a problem with?

wachterjohannes:
maybe @berndhep has one

berndhep:
@wachterjohannes Where can I test it?

wachterjohannes:
@danrot can we deploy it somewhere? or should we test it locally on my machine?

danrot:
@wachterjohannes Can you test it on your machine? Less effort for me 😄

danrot:
The buttons in the overlay are not correct.

wachterjohannes:
@danrot fixed

PR #661, opened 2016-06-14
Added second dot in column navigation for draft

danrot:
This PR adds a second dot in green to the column navigation, indicating that there exists a published version (only shown in combination with a draft). Required for https://github.com/sulu/sulu/pull/2361


chirimoya:
ACK

PR #662, opened 2016-06-17
Added select field to editable inputs

wachterjohannes:
This PR adds the possibility to edit select-fields in the table.


scss/modules/datagrid.scss
```diff
@@ -90,6 +90,10 @@
             height: $inputHeight;
             display: none;
         }
+        select.editable-input {
```
alexander-schranz: add empty line before

wachterjohannes: fixed

PR #664, opened 2016-06-27
collapse dropdown button without visible items in list

danrot:
Dependent on https://github.com/massiveart/husky/pull/661

Makes a dropdown button in the toolbar list a normal button, as soon as all its dropdown items are hidden. Talking about this save button:
image

When all the dropdown items under this button are hidden, it will be a normal button:
image


husky_services/util.js
```diff
@@ -7,6 +7,12 @@
  * with this source code in the file LICENSE.
  */
 
+require.config({
+    paths: {
+        'sprintf': 'bower_components/sprintf/sprintf'
```
danrot: This is necessary here and in the grunt config in order to work correctly...

PR #665, opened 2016-06-29
Wider overlay container

akira28:
Related to Sulu PR https://github.com/sulu/sulu/pull/2514 and issue https://github.com/sulu/sulu/issues/2104

### DIFF
--- a/scss/variables.scss
+++ b/scss/variables.scss
@@ -111,4 +111,4 @@ $navigationHoverColor: #191D22;
 // Overlay
 // -------------------------
 $overlayContainerWidth: 580px;
-$overlayContainerWidthExtraWide: 700px;
+$overlayContainerWidthExtraWide: 750px;


scss/variables.scss
```diff
@@ -111,4 +111,4 @@ $navigationHoverColor: #191D22;
 // Overlay
 // -------------------------
 $overlayContainerWidth: 580px;
-$overlayContainerWidthExtraWide: 700px;
+$overlayContainerWidthExtraWide: 750px;
```
wachterjohannes: we should introduce another skin for that (:

danrot:
Do we need that change now? The linked PR has already been merged without that.

akira28:
Is a styling issue, it depends on what you think about it, is not critical I suppose

akira28:
This is how it looks without the change:

screenshot 2016-07-20 14 46 06

danrot:
What's the problem here? Don't you like that 3 columns don't fit the width?

akira28:
Exactly :D

danrot:
@akira28 We just decided to make this overlay responsive, so that it takes as much as space as it has available. So we won't merge this PR, but you'll see that improvement quite soon on the develop branch :-)

PR #675, opened 2016-07-06
made sure data navigation only loads once

marcelmoosbrugger:


husky_components/navigation/main.js
```diff
@@ -982,6 +992,8 @@ define(function() {
          * Remove data navigation
          */
         removeDataNavigation: function() {
+            if (!!this.dataNavigationIsLoading) return;
```
danrot: Why is that necessary? Shouldn't it be possible to remove the data navigation when it is currently loading?

marcelmoosbrugger:
Closed in favor of https://github.com/massiveart/husky/pull/685

PR #679, opened 2016-07-07
[HOTFIX] Globalize: Fixed loading correct file

trickreich:
Globalize: Fixed loading correct file when culture name includes country


marcelmoosbrugger:
review passed

PR #698, opened 2016-07-27
Created published icons for itembox

marcelmoosbrugger:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| Enhancement? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | none
| Related issues/PRs | none
| License | MIT
| Documentation PR | none

#### What's in this PR?

The itembox got the ability to render yellow and green/yellow icons.


husky_extensions/itembox.js
```diff
@@ -32,6 +32,8 @@ define(function() {
             hideSelectedCounter: false,
             defaultDisplayOption: 'top',
             actionIcon: 'fa-plus-circle',
+            publishedStateName: 'publishedState',
+            publishedName: 'published',
```
danrot: This way you make the `itembox` dependent on the published functionality. Shouldn't it be quite easy to move these changes to the concrete itembox implementations?

husky_extensions/itembox.js
```diff
@@ -490,6 +507,31 @@ define(function() {
...
+                        this.sandbox.translate(this.options.translations.unpublished)
+                    );
+                } else if (!itemData[this.options.publishedStateName]) {
+                    icons = templates.icons.notPublishedState(
+                        this.sandbox.translate(this.options.translations.unpublished)
+                    );
+                }
+
+                return icons;
+            },
```
danrot: I've seen that you are already overriding that in sulu. Why not making that part of the template which is overriden by the sulu component?

husky_extensions/itembox.js
marcelmoosbrugger: I could use generic icon names here and override them in the internal link component.

husky_extensions/itembox.js
marcelmoosbrugger: You surely mean the code in the smart-content. The thing is that the smart-content component does not use the itembox. This is an open issue.

danrot: That's a separate issue for sure, but it also applies to internal links. This problem is IMO not solved by using generic icon names, the icons should not be here at all. Why can't we use the `getItemContent` function in the `itembox` in the internal links to include the icons?

marcelmoosbrugger: Ok. I'll move to the internal-links

husky_extensions/itembox.js
```diff
@@ -49,7 +49,8 @@ define(function() {
                 viewLess: 'public.view-less',
                 of: 'public.of',
                 visible: 'public.visible',
-                elementsSelected: 'public.elements-selected'
+                elementsSelected: 'public.elements-selected',
+                unpublished: 'public.unpublished'
```
danrot: Should also be removed, right?

scss/modules/column-navigation.scss
```diff
@@ -124,24 +122,7 @@ $publishedColor: $green;
...
-                    background-color: $infoColor;
-                }
-
-                .not-published {
-                    background-color: $publishedColor;
-                }
-
-                .not-published + .not-published-state {
+                .not-published-icon + .not-published-state-icon {
                     margin-left: -11px;
```
danrot: Why is this separate from the value in `custom-icons.scss`?

marcelmoosbrugger: The column-navigation defines a default space between its icons. Therfore to make the two icons (yellow and green) overlap, the margin-left property needs to be overridden.

PR #699, opened 2016-07-28
Made url comparison in navigation use url parts an not characters

marcelmoosbrugger:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | none
| Related issues/PRs | none
| License | MIT
| Documentation PR | none

#### What's in this PR?

The navigation had problems with selecting the right navigation item, when it was forced to update. This PR fixes this issue.


husky_services/util.js
```diff
@@ -314,36 +314,59 @@ define(['sprintf'], function(sprintf) {
...
+
+            /**
+             * Calculates the number of matches parts of the first array
+             * have with parts of the second array
+             *
+             * @param {Array} parts1
+             * @param {Array} parts2
+             * @returns {number}
+             */
+            getSuccessfulMatches = function(parts1, parts2) {
```
danrot: Does this method have to be a closure in another function? This could easily produce memory leaks. Can't you put it outside? You probably can, because this method does not touch anything outside of itself.

PR #700, opened 2016-07-29
added datagrid get-records and delete-records event

nnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | https://github.com/sulu/sulu/pull/2704
| License | MIT
| Documentation PR | -

#### What's in this PR?

This PR adds a RECORDS_REMOVE and RECORDS_GET event in the datagrid component.
Also it fixes a small bug which occured when trying to select all items in the table-decorator of the datatagrid.


husky_components/datagrid/decorators/table-view.js
```diff
@@ -1424,7 +1424,7 @@ define(function() {
          * @param event {Object} the event object
          */
         allSelectItemChangeHandler: function(event) {
-            if (this.data.total == 0) {
+            if (this.data.embedded.length == 0) {
```
danrot: I guess this only changed because you are not returning a `ListRepresentation` as you've written in https://github.com/sulu/sulu/pull/2704#discussion_r72930210, right? Maybe we should return it there, and leave this the same, so that it also works with pagination?

husky_components/datagrid/main.js
```diff
@@ -1691,6 +1712,19 @@
             },
 
             /**
+             * Handles the row remove event
+             */
+            removeRecordsHandler: function(recordIds) {
+                if (!!this.gridViews[this.viewId].removeRecord && !!recordIds) {
```
danrot: Reverse logic and return early.

husky_components/datagrid/decorators/table-view.js
nnatter: No, the problem here is, that the select-all toggle does not work, if there were no server-side list items at the point of last load. If we add/remove elements to the datagrid locally, `this.data.total` will not change and therefore the toggle will not do anything.

Returning a `ListRepresentation` will fix the toggle only after saving, but not before.
I think this should work with pagination, as long we do no have empty pages?

CHANGELOG.md
```diff
@@ -1,11 +1,15 @@
...
+* dev-develop
+    * ENHANCEMENT #700 added datagrid get-records and delete-records event
+
 * 0.24.0 (2016-08-08)
     * ENHANCEMENT #698 Added responsive-width skin in overlay component
     * ENHANCEMENT #702 Implemented error handling on column-navigation load
     * BUGFIX      #699 Made url comparison in navigation use url parts an not characters
     * ENHANCEMENT #698 Created published icons for itembox
     * BUGFIX      #701 Prefixed the slide class of the search component
+
```
danrot: Please remove this empty line

PR #702, opened 2016-08-03
Implemented error handling on column-navigation load

marcelmoosbrugger:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| Enhancement? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | none
| License | MIT
| Documentation PR | -

#### What's in this PR?

The url option passed to the column-navigation can now also be a function. The column-navigation tries to load the initial content a second time from another url, if fetching the data with a selected node fails.


husky_components/column-navigation/main.js
```diff
@@ -409,14 +409,37 @@ define(function() {
...
             }
         },
 
         /**
+         * Returns the url which is used to initially load the column-navigation
+         *
+         * @param {Boolean} errorOccured
+         * @returns {String} an url to load the column tree
+         */
+        getInitialUrl: function(errorOccured) {
```
danrot: Can we introduce a `fallbackUrl` instead? Would be easier to understand, and is more flexible.

PR #706, opened 2016-08-08
Created tiles view for datagrid

marcelmoosbrugger:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | none
| License | MIT
| Documentation PR | -

#### What's in this PR?

This PR contains a new view decorator for the datagrid.


apimocks/datagrid/tiles.json
```diff
@@ -0,0 +1,74 @@
...
+        "self": {
+            "href": "http://husky.lo:7878/admin/api/datagrid/tiles"
+        }
+    },
+    "_embedded": {
+        "items": [
+            {
+                "id": "1",
+                "title": "Folder title 1",
+                "elements": 100
```
danrot: In other APIs we call this field `total`, indicates to me better that this is a number.

apimocks/datagrid/tiles.json
```diff
@@ -0,0 +1,74 @@
...
+                "elements": 210
+            },
+            {
+                "id": "12",
+                "title": "Folder title 12",
+                "elements": 220
+            }
+        ]
+    },
+    "total": 12,
```
danrot: Like here 😄

husky_components/datagrid/decorators/tiles-view.js
```diff
@@ -0,0 +1,226 @@
...
+
+            return description.trim();
+        },
+
+        /**
+         * Destroys the view
+         */
+        destroy: function() {
+            this.$el.remove();
+            this.setVariables();
```
danrot: Why do we need that call here? Asking because we could inline the method in the `initialize` method otherwise.

husky_components/datagrid/decorators/tiles-view.js
```diff
@@ -0,0 +1,226 @@
...
+            this.$el.append(this.$tiles[record.id]);
+        },
+
+        /**
+         * Binds dom events for a single tile
+         *
+         * @param {Object} tileId The id of the tile to bind the events for
+         */
+        bindTileDomEvents: function(tileId) {
+            this.$tiles[tileId].on('click', function() {
```
danrot: Wouldn't it be easier and more efficient to bind this event once with a filter, instead for every single tile?

scss/modules/tiles.scss
```diff
@@ -0,0 +1,72 @@
...
+        display: inline-block;
+        margin: 0 $spacing $spacing 0;
+        cursor: pointer;
+        text-align: center;
+        vertical-align: middle;
+        border-radius: $borderRadiusSmall;
+    }
+
+    .tile {
+        border-radius: $borderRadiusSmall;
```
danrot: Isn't that the same as defined in the previous selector? Why do you need a second selector for the tile solely at all? Does it look different?

danrot: Ok, just saw that they look quite different... but they seem to share the `border-radius`, so you can remove at least that one 😉

husky_components/datagrid/decorators/tiles-view.js
marcelmoosbrugger: The properties of the decorator need to be reseted, when the datagrid calls the destroy method. We can't inline it in the initialize method, as it get's only called at the very beginning. The render and destroy methods, however, get called multiple times.

danrot: And why do they need to be reset?

apimocks/datagrid/tiles.json
```diff
@@ -0,0 +1,74 @@
...
+            },
+            {
+                "id": "11",
+                "title": "Folder title 11",
+                "countElements": 210
+            },
+            {
+                "id": "12",
+                "title": "Folder title 12",
+                "countElements": 220
```
danrot: If you call it like that I would say `mediaCount`. Having `count` first makes it sound like a function, and if you only call it elements I would suspect it to also include sub collections.

PR #707, opened 2016-08-08
Added translations to titles in matrix

ReneZumtobel:
Fixed the bug with the languages in setting, user roles.


danrot:
@chirimoya How should we handle the dutch translations now? Should I try to get crowdin running? Then we could omit french and dutch, and say these are community translated, and pull the changes from time to time.

PR #712, opened 2016-08-11
Removed data-navigation from navigation component

marcelmoosbrugger:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| Enhancement? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | https://github.com/massiveart/husky/pull/706
| License | MIT
| Documentation PR | -

#### What's in this PR?

see title.

#### Why?

The data-navigation was not needed anymore in the navigation component.


husky_components/datagrid/main.js
```diff
@@ -1506,22 +1508,34 @@
...
+                    return;
+                }
+
+                this.show = !this.show;
+                this.gridViews[this.viewId].showSelected(this.show);
+                if (!!this.show && !!this.paginations[this.paginationId]) {
+                    this.paginations[this.paginationId].destroy();
+                } else {
+                    this.rerenderPagination();
+                }
```
marcelmoosbrugger: The destroying of the pagination was added because the datagrid showed weird behavior when trying to paginate in the mode where only the selected records are shown. Moreover the `infinite-scroll` pagination had a pretty critical issue in this case, because as soon as the `show only selected` button was clicked, it thought that the bottom of the page has been reached and started paginating.

danrot:
Shouldn't we also delete `data-navigation` fromthe `husky_components` folder?

marcelmoosbrugger:
@danrot It is still used in the media-selection

PR #713, opened 2016-08-16
Removed overlay from dropzone and implemented own overlay-style for dropzone

marcelmoosbrugger:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| Enhancement? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs |  -
| License | MIT
| Documentation PR | -

#### What's in this PR?

The dropzone can be displayed in an overlay-like visualization. Before this PR the dropzone did this by using the overlay component. This PR removed the usage of the overlay component from the dropzone and implements the desired behavior directly within the dropzone.

#### Why?

There's a new design for the dropzone overlay, which differs tremendously from the look of the overlay component. Moreover the behavior of the dropzone overlay is different from the one implemented in the overlay component.


UPGRADE.md
```diff
@@ -1,5 +1,19 @@
 # Upgrade
 
+## dev-develop
+
+### Lock and Unlocking the dropzone overlay.
```
danrot: Remove the dot at the end.

danrot: Is that something a user really has to know?

scss/modules/dropzone.scss
```diff
@@ -3,285 +3,339 @@
...
+            .icon {
+                height: 25px;
+                color: $grayLight;
+                font-size: 25px;
+                text-align: center;
+                display: block;
+                animation: ROTATE-ANIMATION 1s linear infinite;
+                -webkit-animation: ROTATE-ANIMATION 1s linear infinite;
+                -moz-animation: ROTATE-ANIMATION 1s linear infinite;
+                -o-animation: ROTATE-ANIMATION 1s linear infinite;
```
danrot: Isn't there a mixin for that? Like for the transitions?

scss/modules/dropzone.scss
```diff
@@ -3,285 +3,339 @@
...
-        line-height: 25px;
-      }
-      .title {
-        font-size: $dropzoneSmallFontSize;
-        line-height: $dropzoneSmallFontSize;
-      }
-    }
-    .upload-items {
-      height: $smallHeight;
+            &.dz-success {
```
danrot: no abbreviations

scss/modules/dropzone.scss
```diff
@@ -3,285 +3,339 @@
...
+                    }
+                }
             }
+        }
 
-            .progress {
-              width: 80px;
-              height: 8px;
-              border-radius: 4px;
+        &:hover, &.dz-drag-hover, &.dropped {
```
danrot: Also no abbreviation `dz`

scss/modules/dropzone.scss
marcelmoosbrugger: That's a class provided by the dropzone plugin (http://www.dropzonejs.com/). We can't really influence that.

danrot: ok, then forget about it :)

dist/scss/mixins.scss
```diff
@@ -49,6 +49,14 @@
   -webkit-transform:rotate($angle); /* Opera, Chrome, and Safari */
 }
 
+@mixin animation($str) {
+    -webkit-animation: #{$str};
+    -moz-animation: #{$str};
+    -ms-animation: #{$str};
+    -o-animation: #{$str};
+    animation: #{$str};
+}
```
danrot: Do we have to implement that on our own? Can't we use the [compass mixin](http://compass-style.org/reference/compass/css3/animation/)?

husky_components/dropzone/main.js
```diff
@@ -284,22 +284,26 @@ define([], function() {
                         this.addFiles(event.originalEvent.dataTransfer.files);
                     }
                 }.bind(this));
+                this.$dropzone.on('click', function() {
+                    if (!!this.options.cancelUploadOnOverlayClose) {
```
danrot: Why have you moved this one out of the `closeOverlay` function? Is there any situation in where reopening the dropzone should not drop all the files?

marcelmoosbrugger: That's what the configuration is about. The upload's should only be canceled if the `cancelUploadOnOverlayClose` is set and the overlay is closed with a click by the user. If this piece of code was in the `closeOverlay` method, the uploaded files would also get canceled when they have been successfully uploaded (because the overlay closes after that). From the server perspective everything is fine, however the dropzone in the frontend views the uploaded files as canceled and triggers, that no files have been uploaded.

PR #715, opened 2016-08-17
improved lazy-loading of children in datagrid

nnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | https://github.com/sulu/sulu/pull/2823
| License | MIT
| Documentation PR | -

#### What's in this PR?

This PR implements three improvements for handling lazy loading of children in the husky datagrid.

#### Why?

1. replace expand-icon with loader when loading children in datagrid
While loading childred, the expand icon of the table row is replaced with a loader to signal that loading is in progress.

2. append currently selected items to request when loading data in datagrid
Lazy loading children leads to problems if a child-item should be preselected. Preselected items are marked as selected before the first load now. The currently selected items are appended to the request when loading data, to signal the server, that the path to this items should be expanded in the response.

3. display effective limit in dropdown pagination component
Until now, the number of displayed children was displayed as page-size in the dropdown pagination component. When displaying nested data, this no longer makes sense, because pagination should only affect the root level of the data. With this PR, the dropdown pagination component displays the effective limit which is sent to the server as page size.


husky_components/datagrid/main.js
```diff
@@ -1007,6 +1005,13 @@
              */
             load: function(params) {
                 this.currentUrl = this.getUrl(params);
+
+                if (!!this.getSelectedItemIds().length) {
+                    var delimiter = (this.currentUrl.indexOf('?') === -1) ? '?' : "&",
+                        selectedString = this.getSelectedItemIds().join(',');
+                    this.currentUrl = this.currentUrl + delimiter + 'expand=' + selectedString;
```
nnatter: im not sure, if `expand` is the right name for this parameter

husky_components/datagrid/main.js
```diff
@@ -1007,6 +1005,13 @@
              */
             load: function(params) {
                 this.currentUrl = this.getUrl(params);
+
+                if (!!this.getSelectedItemIds().length) {
```
nnatter: should we invent a datagrid option to enable/disable this behaviour?

scss/modules/datagrid.scss
```diff
@@ -194,6 +194,7 @@
 
         .child-wrapper {
             .toggle-icon {
+                display: inline-block;
```
nnatter: this prevents line breaking when the loader is displayed

husky_components/datagrid/main.js
```diff
@@ -1007,6 +1005,10 @@
              */
             load: function(params) {
                 this.currentUrl = this.getUrl(params);
+
+                if (!params.data) params.data = {};
+                params.data['expand'] = this.getSelectedItemIds().join(',');
```
nnatter: im not sure, if `expand` is the right name for this parameter

nnatter: should we invent a datagrid option to enable/disable this behaviour?

danrot: Can't think of any better name at the moment... I'd say implementing an option is not necessary and makes things only more complicated.

husky_components/datagrid/main.js
```diff
@@ -2258,18 +2262,25 @@
...
                     var template = this.sandbox.uritemplate.parse(this.data.links.children.href),
                         url = this.sandbox.uritemplate.expand(template, {parentId: recordId});
 
                     this.sandbox.util.load(this.getUrl({url: url}))
                         .then(function(response) {
                             this.addRecordsHandler(response._embedded[this.options.resultKey]);
+                            def.resolve();
                         }.bind(this))
                         .fail(function(status, error) {
                             this.sandbox.logger.error(status, error);
```
danrot: Shouldn't the `def` be failing here?

husky_components/datagrid/main.js
```diff
@@ -2258,18 +2262,25 @@
              * @param recordId {Number|String}
              */
             loadChildren: function(recordId) {
+                var def = this.sandbox.data.deferred();
```
danrot: `deferred` - we don't use abbreviations

PR #717, opened 2016-08-18
Bugfix: Matrix select all none

ReneZumtobel:
| Q | A |
| --- | --- |
| Bug fix? | yes |
| New feature? | no |
| BC breaks? | no |
| Deprecations? | no |
| Fixed tickets | fixes #issuenum |
| Related issues/PRs | https://github.com/sulu/sulu/pull/2834 |
| License | MIT |
#### What's in this PR?

Fixed bug with select all/ select none problems and changed design


husky_components/matrix/main.js
```diff
@@ -187,9 +189,9 @@ define(function() {
         prepareTableRow: function($tbody, rowCount) {
             var $tr = this.sandbox.dom.createElement('<tr/>'),
                 $tdHead = this.sandbox.dom.createElement('<td class="section"/>'),
-                $tdValue = this.sandbox.dom.createElement('<td class="value"/>'),
-                $tdAll = this.sandbox.dom.createElement('<td class="all"/>'),
-                allActive;
+                $tdValue = this.sandbox.dom.createElement('<td class="value"></td>'),
+                $tdAll = $tdValue,
```
danrot: How does that make sense? I am quite sure these two elements should not be the same.

husky_components/matrix/main.js
```diff
@@ -187,9 +189,9 @@ define(function() {
...
+                allInActive;
```
danrot: `allInactive`

husky_components/matrix/main.js
```diff
@@ -205,7 +207,7 @@ define(function() {
 
             // flag for checking if every flag is true
```
danrot: Update comment

husky_components/matrix/main.js
```diff
@@ -215,21 +217,20 @@ define(function() {
             this.sandbox.dom.append($tr, $tdValue);
 
             //add all link
-            this.sandbox.dom.html(
-                $tdAll,
-                [
+                $tdAll.append(
+                [   '<div class="all">',
```
danrot: Why does this have to be wrapped in a div? The `display` value is set to`inline-block`, this could also be set on the span. And does it have to `inline-block`? Why can't it be `inline`, the default value of the `span` element?

husky_components/matrix/main.js
ReneZumtobel: that´s because the all button is now part of tdAll and not a separate td

husky_components/matrix/main.js
```diff
@@ -32,7 +32,7 @@ define(function() {
                 this.toggleIcon.bind(this),
                 'tbody > tr > td.value > span[class^="fa-"]'
             );
-            this.sandbox.dom.on(this.$element, 'click', this.toggleRow.bind(this), 'tbody > tr > td:last-child');
+            this.sandbox.dom.on(this.$element, 'click', this.toggleRow.bind(this), 'tbody > tr > td > span:last-of-type');
```
danrot: Shouldn't it still be `td:last-child`? I think this would also merge the last span of other `td`s

scss/modules/matrix.scss
```diff
@@ -24,6 +24,17 @@
...
+
+            > span:last-of-type{
+                display: none;
+            }
+        }
+
+        &.value:hover {
+            > span:last-of-type{
+                display: inline;
+                padding-left: 3px;
```
danrot: Would move that after the `display:none` line. Then there is less happening on hover, which might be a little bit more performant.

husky_components/matrix/main.js
```diff
@@ -42,15 +42,16 @@ define(function() {
...
         toggleIcon: function(event) {
             var $target = event.currentTarget,
-                $tr = this.sandbox.dom.parent($target),
-                $allTargets = this.sandbox.dom.find('span[class^="fa-"]', $tr),
+                $td = this.sandbox.dom.parent($target),
+                $tr = this.sandbox.dom.parent($td),
+                $allTargets = this.sandbox.dom.find('span[class^="fa-"]', $td),
                 $activeTargets,
-                $link = this.sandbox.dom.find('td:last-child span', $tr);
+                $link = this.sandbox.dom.find('td:last-child > span:last-of-type', $tr);
```
danrot: Don't you also want to use `last-child` here instead of `last-of-type`? Not that we are running in some strange behavior, because we are using another selector above.

husky_components/matrix/main.js
```diff
@@ -65,12 +66,12 @@ define(function() {
         },
 
         toggleRow: function(event) {
-            var $tr = this.sandbox.dom.parent(event.currentTarget),
+            var $tr = this.sandbox.dom.parent(this.sandbox.dom.parent(event.currentTarget)),
                 $targets = this.sandbox.dom.find('span[class^="fa-"]', $tr),
                 $activeTargets = this.sandbox.dom.find('span[class^="fa-"].' + activeClass, $tr),
-                $link = this.sandbox.dom.find('td:last-child span', $tr), activated;
+                $link = this.sandbox.dom.find('td:last-child > span:last-of-type', $tr), activated;
```
danrot: This selector is already used in my previous variable. How about extracting it to a separate variable?

husky_components/matrix/main.js
```diff
@@ -95,6 +96,8 @@ define(function() {
 
             // emit events for communication with the outside
             this.sandbox.dom.each($trs, function(key, tr) {
+                var $link = this.sandbox.dom.find('td:last-child > span:last-of-type', tr);
```
danrot: Same as above

husky_components/matrix/main.js
```diff
@@ -110,6 +113,8 @@ define(function() {
 
             // emit events for communication with the outsite
             this.sandbox.dom.each($trs, function(key, tr) {
+                var $link = this.sandbox.dom.find('td:last-child > span:last-of-type', tr);
```
danrot: same as above

husky_components/matrix/main.js
```diff
@@ -215,21 +216,18 @@ define(function() {
             this.sandbox.dom.append($tr, $tdValue);
 
             //add all link
-            this.sandbox.dom.html(
-                $tdAll,
-                [
-                    '<span class="pointer">',
-                    (!!allActive) ? this.options.captions.none : this.options.captions.all,
+                $tdValue.append(
+                [   '<span class="all pointer">',
```
danrot: new line for the span

husky_components/matrix/main.js
```diff
@@ -215,21 +216,18 @@ define(function() {
...
+                    (!allInactive) ? this.options.captions.none : this.options.captions.all,
                     '</span>'
-                ].join('')
-            );
-            this.sandbox.dom.append($tr, $tdAll);
+                ].join(''));
```
danrot: last `);` should go into own line

scss/modules/matrix.scss
```diff
@@ -24,6 +24,17 @@
...
             white-space: nowrap;
+
+            > span:last-of-type{
+                display: none;
+                padding-left: 3px;
+            }
+        }
+
+        &.value:hover {
+            > span:last-of-type{
```
danrot: Also use `last-child` for both css selectors in here.

husky_components/matrix/main.js
```diff
@@ -11,7 +11,8 @@ define(function() {
 
     'use strict';
 
-    var activeClass = 'is-active';
+    var activeClass = 'is-active',
+        currentSelectButton = 'td:last-child > span:last-child';
```
danrot: That reads like a selector, not a button, so the naming is not really appropriate. And why `current`? Isn't that selector valid for all the buttons?

husky_components/matrix/main.js
```diff
@@ -32,7 +33,7 @@ define(function() {
                 this.toggleIcon.bind(this),
                 'tbody > tr > td.value > span[class^="fa-"]'
             );
-            this.sandbox.dom.on(this.$element, 'click', this.toggleRow.bind(this), 'tbody > tr > td:last-child');
+            this.sandbox.dom.on(this.$element, 'click', this.toggleRow.bind(this), 'tbody > tr > td:last-child > span:last-child');
```
danrot: Why not using the selector defined above?

husky_components/matrix/main.js
```diff
@@ -203,9 +205,9 @@ define(function() {
             this.sandbox.dom.data($tdHead, 'section', this.options.values.vertical[rowCount]);
             this.sandbox.dom.append($tr, $tdHead);
 
-            // flag for checking if every flag is true
+            // flag for checking if every flag is false
             // insert values of matrix
-            allActive = this.prepareTableColumn(
+            allInactive = this.prepareTableColumn(
```
danrot: Can't we leave that as `allActive`? Names phrased in a positive way are easier to understand.

ReneZumtobel: Actually because of the use case I don´t think `allActive` would be the better choice because I check if there is only a single one selected. That means `!allActive` is not the same as `allInactive`.

husky_components/matrix/main.js
```diff
@@ -11,7 +11,8 @@ define(function() {
 
     'use strict';
 
-    var activeClass = 'is-active';
+    var activeClass = 'is-active',
+        selectButton = 'td:last-child > span:last-child';
```
danrot: I have already written before that this variable is not representing a button, so rename it to what it really is: `selectButtonSelector`

husky_components/matrix/main.js
danrot: Ok, leave it

danrot: Or rename it to `someActive`? Would be the actual positive variant of this name.

husky_components/matrix/main.js
```diff
@@ -215,21 +217,19 @@ define(function() {
...
+                    ].join('')
+                );
+            this.sandbox.dom.append($tr, $tdValue);
 
             this.sandbox.dom.append($tbody, $tr);
         },
 
         prepareTableColumn: function($tdValue, columnData, section, value) {
-            var allActive = true,
+            var allInactive = true,
```
danrot: Should also be `someActive`

husky_components/matrix/main.js
```diff
@@ -215,21 +217,19 @@ define(function() {
...
-                [
-                    '<span class="pointer">',
-                    (!!allActive) ? this.options.captions.none : this.options.captions.all,
-                    '</span>'
-                ].join('')
-            );
-            this.sandbox.dom.append($tr, $tdAll);
+                $tdValue.append(
+                    [   '<span class="all pointer">',
+                        (someActive) ? this.options.captions.none : this.options.captions.all,
```
danrot: `!!someActive`

PR #723, opened 2016-08-24
Added description class to itembox and made content take up the whole width

marcelmoosbrugger:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

The itembox got the ability to render a description. Moreover the markup was slightly changed, such that the container in which the content of the subtypes is rendered takes up the whole width.

### Why?
When the container in which the content is rendered takes up all remaining width, more advanced positioning is possible than if everything is just inline.


husky_extensions/itembox.js
```diff
@@ -110,7 +110,7 @@ define(function() {
                     '<li data-id="', id, '">',
                     !!this.options.sortable ? '    <span class="fa-ellipsis-v icon move"></span>' : '',
                     '    <span class="num"></span>',
-                    content,
+                    '   <div class="item-content">' + content + '</div>',
```
danrot: Indentation not correct

scss/modules/content-types.scss
```diff
@@ -354,28 +358,48 @@
                 .num {
                     color: $grayLight;
                     padding-right: 15px;
+                    display: table-cell;
+                    width: 1%; // make cell fit content
```
danrot: Any better explanation why this is necessary?

marcelmoosbrugger: What do you want me to describe in more detail? The only reason for the `1%`, is that the table-cell should fit its content. I don't really know what to add to this. This a common pattern and has to do with the table-layout and how the browser renders it.

danrot: But why `1%`? Seems like an odd choice to me. Doesn't it also work with `1px`?

marcelmoosbrugger: It works with every width that is smaller than the content in the cell but still greater than 0. Do you prefer pixels? How do you think, does using pixels improve the component?

danrot: Well, the difference for me is that already one letter is probably longer than 1px, this cannot be guaranteed for percentages.

PR #725, opened 2016-08-25
Added locale to pagination request of data-navigation

marcelmoosbrugger:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | - 
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

The locale-parameter was not sent with the pagination request, although the API demanded it. This PR fixes this issue.


husky_components/data-navigation/main.js
```diff
@@ -338,17 +338,18 @@ define([
...
-                    .then(function(data) {
-                        var children = data._embedded[this.options.resultKey] || [];
-                        this.data.children = this.data.children.concat(children);
-                        this.data.hasNextPage = this.page < data.pages;
-                        this.currentView.append(children, this.options);
-
-                        this.hideLoader();
-                        def.resolve();
-                    }.bind(this));
+                this.page += 1;
```
danrot: Why this change? And would prefer `++this.page`

marcelmoosbrugger: It's just a matter of taste. I'm not a fan of "++" and "--" as they have side-effects. In my opinion "+=" makes makes it more clear that there's an assignment happening.

PR #727, opened 2016-08-30
Added isTabChangeAllowed option to tab component

ReneZumtobel:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | yes |
| BC breaks? | no |
| Deprecations? | no |
| Fixed tickets | fixes |
| Related issues/PRs | https://github.com/sulu/sulu/pull/2884 |
| License | MIT |
#### What's in this PR?

Added isTabChangeAllowed option to tab component.


demos/tabs/index.html
```diff
@@ -76,6 +80,29 @@
                         app.logger.log('triggered select item via pre select (manually)');
                         app.sandbox.emit('husky.tabs.contacts.item.clicked', event);
                     });
+
+                    var isTabChangeAllowed = function () {
+                        var allowed = $.Deferred();
+                        if (Math.floor(Math.random() * 2) >= 1) {
```
danrot: Why not simply creating two different tabs? This way you would have to manually try to reload until you get the desire behavior to test.

husky_components/tabs/main.js
```diff
@@ -15,6 +15,7 @@
  *      - data: if no url is provided
  *      - selected: the item that's selected on initialize
  *      - instanceName - enables custom events (in case of multiple tabs on one page)
+ *      - tabChangeAllowed: funtion that checks if tab change is allowed
```
danrot: `function`

husky_components/tabs/main.js
```diff
@@ -171,27 +173,39 @@ define(function() {
...
-                        item.callback.call(this, item);
-                    } else if (!!this.options.callback && typeof this.options.callback === 'function') {
-                        this.options.callback.call(this, item);
-                    } else {
-                        triggerSelectEvent.call(this, item);
-                    }
+                var tabChangeAllowed = false;
+
+                if (typeof this.options.tabChangeAllowed === 'function') {
+                    this.options.tabChangeAllowed().then(function () {
```
danrot: Is there a reason we are using defereds here?

husky_components/tabs/main.js
```diff
@@ -171,27 +173,39 @@ define(function() {
...
                 }
             } else {
                 return false;
             }
         },
 
+        changeTab = function (event) {
+            var item = this.items[this.sandbox.dom.data(event.currentTarget, 'id')];
+
+            if (!!item) {
```
danrot: Reverse logic:
```javascript
if (!item) {
    return;
}
```

husky_components/tabs/main.js
```diff
@@ -54,7 +55,8 @@ define(function() {
             preSelectEvent: {
                 enabled: false,
                 triggerSelectItem: true
-            }
+            },
+            tabChangeAllowed: null
```
danrot: Does this component have to know about this at all? Isn't it the job of the component using the tabs to determine if the switch should really happen or not?

danrot: Another solution would be to only pass a function which returns true or false, and check its return value in here. Then we wouldn't have to use deferreds here, which makes it easier to use.

danrot: @ReneZumtobel That would have been the crucial part: Moving that logic to the other component. However, it might be a better idea to change this to a callback which tells the tabs only if the change is allowed or not, and create the overlay in here. Would have the advantage that we could easily reuse it for other forms containing tabs.

danrot (changes requested):

PR #730, opened 2016-09-07
Overlay and label adaptions

marcelmoosbrugger:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | https://github.com/sulu/sulu/pull/2908
| License | MIT
| Documentation PR | -

#### What's in this PR?

The new cropping feature in Sulu made some changes in husky necessary. These are introduced by this PR.
- The overlay got a so called "panel" on the left of the language-changer. Through the options the panel can be filled with content
- The overlay backdrop closed the component on mouseup. To solve this issue, the dom structure of the overlay had to be slightly changed.
- The label got a "vanish" event, to be able to make a label disappear from outside the component.
- The floats within the select-component were replaced by "inline-block" due to weird rendering bugs.


dist/scss/variables.scss
```diff
@@ -111,6 +111,6 @@ $navigationHoverColor: #191D22;
 // Overlay
 // -------------------------
 $overlayContainerWidthSmall: 580px;
-$overlayContainerWidthMedium: 700px;
+$overlayContainerWidthMedium: 860px;
```
danrot: Why is that necessary? There is almost no difference anymore between medium and large...

danrot: And this also has an impact on already existing overlays, doesn't it?

husky_components/overlay/main.js
```diff
@@ -36,12 +35,15 @@
  * @params {Function} [options.slides[].closeCallback] @deprecated Use 'cancelCallback' instead
  * @params {Function} [options.slides[].cancelCallback] callback which gets executed after the overlay gets canceled
  * @params {Function} [options.slides[].okCallback] callback which gets executed after the overlay gets submitted
+ * @params {Boolean} [options.slides[].displayHeader] callback which gets executed after the overlay gets submitted
```
danrot: Does this line make sense? Is `displayHeader` the type of callback described here?

dist/scss/variables.scss
marcelmoosbrugger: The decision of the design section was to make the medium width larger.
That's right, this has an impact on already existing overlays.

danrot: But then we also have to adapt [this line](https://github.com/massiveart/husky/blob/develop/scss/modules/overlay.scss#L41), don't we? And maybe some more, but that's the only one I found so far :)

marcelmoosbrugger: The breakpoints should be fixed now. As far as I know the breakpoints only affected the media-selection.

PR #737, opened 2016-09-28
added possibility to define overlay spacing for each slide

danrot:
The `contentSpacing` option of the overlay adds a `content-spacing` class, which introduces a padding of 20 pixels. However, this option had an impact on all the slides of the overlay, and it couldn't be set differently for different slides. This PR introduces that possibility.


wachterjohannes:
ACK

PR #738, opened 2016-09-29
removed action icon in datagrid if no icon is defined

danrot:
The datagrid was showing the action icon also if no action icon is defined. In this case only a small blue block without any icon was displayed. This PR fixes that.


wachterjohannes:
ACK

PR #740, opened 2016-09-29
Fixed null values for get-selected in datagrid

wachterjohannes:
If you have selected ids which are not in the list you get null values returned.


husky_components/datagrid/main.js
```diff
@@ -1565,7 +1565,10 @@
                             items = [];
 
                         this.sandbox.util.foreach(ids, function(id) {
-                            items.push(this.getRecordById(id));
+                            var item = this.getRecordById(id);
+                            if (!!item) {
+                                items.push(item);
+                            }
```
danrot: Can't we remove the item here somehow, if it doesn't exist? Because currently the categories are not removed in the smart content overlay, when they are deleted. And there is also no possibility to do so afterwards.

PR #742, opened 2016-10-06
Extracted expression-language

wachterjohannes:
This PR extract the expression-language from tab-component. Used in https://github.com/sulu/sulu/pull/2967


husky_services/expression-language.js
```diff
@@ -0,0 +1,73 @@
...
+    var instance,
+        getInstance = function() {
+            if (instance == null) {
+                instance = new ExpressionLanguage();
+            }
+
+            return instance;
+        };
+
+    function ExpressionLanguage() {
```
danrot: Is `ExpressionLanguage` really a matching term? We are only evaluating based on javascript object and not introducing another kind of language.

husky_services/expression-language.js
```diff
@@ -0,0 +1,73 @@
...
+
+    'use strict';
+
+    var instance,
+        getInstance = function() {
+            if (instance == null) {
+                instance = new ExpressionLanguage();
+            }
+
+            return instance;
```
danrot: A require module is already a singleton, so there is no need to do something like this.

husky_services/logger.js
```diff
@@ -0,0 +1,28 @@
...
+define(function() {
+
+    'use strict';
+
+    var instance = null,
+        getInstance = function() {
+            if (instance == null) {
+                instance = new Logger();
+            }
+            return instance;
```
danrot: Same as above, require modules are already singletons.

husky_services/logger.js
```diff
@@ -0,0 +1,28 @@
...
+        };
+
+    function Logger() {
+    }
+
+    Logger.prototype = window.Husky.logger;
```
danrot: How is the `logger` injected in Husky? Is there no better way than retrieving it from the `window` variable?

wachterjohannes: the logger will be initialized by aura and i have no really way to get it over require. we have done the same with mediator also.

wachterjohannes:
@danrot fixed

.grunt/husky.json
```diff
@@ -32,6 +32,8 @@
             "services/husky/mediator": "husky_services/mediator",
             "services/husky/translator": "husky_services/translator",
             "services/husky/url-validator": "husky_services/url-validator",
+            "services/husky/logger": "husky_services/logger",
+            "services/husky/expression": "husky_services/expression-language",
```
danrot: `expression`

wachterjohannes:
fixed

PR #745, opened 2016-10-20
Fixed clickable ghost pages in column navigation

ReneZumtobel:
| Q | A |
| --- | --- |
| Bug fix? | yes |
| New feature? | no |
| BC breaks? | no |
| Deprecations? | no |
| Fixed tickets | fixes #issuenum |
| Related issues/PRs | https://github.com/sulu/sulu/pull/2997 |
| License | MIT |
#### What's in this PR?

Fixed clickable ghost pages in column navigation.


husky_components/column-navigation/main.js
```diff
@@ -1339,7 +1339,9 @@ define(function() {
          */
         mark: function(id, item) {
             var $element = this.$find('li[data-id="' + id + '"]');
-            if (!!$element.length) {
+
+            // mark only non-ghost pages if showActionButtonOnGhost is disabled
+            if (!!$element.length && (item.type.name !== 'ghost' || this.options.showActionButtonOnGhost)) {
```
danrot: I get an error on this line, when double clicking on an element I should be able to add:

```
husky.js:42880 Uncaught TypeError: Cannot read property 'name' of undefined
```

ReneZumtobel:
@danrot I integrated your comment

husky_components/column-navigation/main.js
```diff
@@ -1344,7 +1344,9 @@ define(function() {
          */
         mark: function(id, item) {
             var $element = this.$find('li[data-id="' + id + '"]');
-            if (!!$element.length) {
+
+            // mark only non-ghost pages if showActionButtonOnGhost is disabled
+            if (!!$element.length && ((typeof item.type === 'undefined' || item.type.name !== 'ghost') || this.options.showActionButtonOnGhost)) {
```
danrot: Usually we just write something like `!!item.type` instead of using `typeof`.

husky_components/column-navigation/main.js
```diff
@@ -1344,7 +1344,9 @@ define(function() {
          */
         mark: function(id, item) {
             var $element = this.$find('li[data-id="' + id + '"]');
-            if (!!$element.length) {
+
+            // mark only non-ghost pages if showActionButtonOnGhost is disabled
+            if (!!$element.length && ((!!item.type || item.type.name !== 'ghost') || this.options.showActionButtonOnGhost)) {
```
danrot: Still getting the `Cannot read property 'name' of undefined` error. Probably because it should be `!!item.type && item.type.name !== 'ghost'` instead of `||`. Not sure if that covers the entire problem though.

husky_components/column-navigation/main.js
```diff
@@ -1344,7 +1344,9 @@ define(function() {
          */
         mark: function(id, item) {
             var $element = this.$find('li[data-id="' + id + '"]');
-            if (!!$element.length) {
+
+            // mark only non-ghost pages if showActionButtonOnGhost is disabled
+            if (!!$element.length && ((!item.type || !!item.type && item.type.name !== 'ghost') || this.options.showActionButtonOnGhost)) {
```
danrot: I think we can get rid of `!!item.type &&` in that case. If `!item.type` is not true, then `item.type` exists and the other statement will work as expected.

PR #749, opened 2016-11-10
Added null check for url-validate

### LINKED ISSUE #747: UrlValidator on null throws uncaught TypeError
Using the `url-input` component with an initial `null` value throws uncaught `TypeError`.
In [line 88 of url-validator.js](https://github.com/massiveart/husky/blob/develop/husky_services/url-validator.js#L88)

`url` should be checked before calling match on it.

wachterjohannes:
Fixes #747


danrot:
Thank you, but please use your own fork for PRs in the future 😉

PR #755, opened 2016-12-06
Added notification badge to tab-component

wachterjohannes:
This PR introduces a notification-badge to tab-component.


danrot:
Would it be a lot of work to extract that into an own component? I know that there is a design for something similar for the dashboard, where it would be needed not in the tabs but in the navigation.

wachterjohannes:
@danrot fixed

PR #756, opened 2016-12-09
Added info class for toolbar

wachterjohannes:
This PR introduces a info-class for toolbar buttons.


scss/modules/datagrid-table-view-light.scss
```diff
@@ -32,12 +32,15 @@
...
+            tr {
+                &.selected {
+                    border-left: none;
+                }
 
-            .row.even, .row.odd {
-                background: transparent;
+                &:nth-child(2n),
+                &:nth-child(2n+1) {
+                    background: transparent !important;
```
danrot: Is this `!important` really necessary?

scss/modules/datagrid-table-view-light.scss
```diff
@@ -32,12 +32,15 @@
...
         tbody {
-            tr.selected {
-                border-left: none;
-            }
+            tr {
+                &.selected {
+                    border-left: none;
+                }
 
-            .row.even, .row.odd {
```
danrot: Shouldn't the class names be deleted somewhere also?

scss/modules/datagrid-table-view-light.scss
```diff
@@ -32,12 +32,15 @@
...
-                background: transparent;
+                &:nth-child(2n),
+                &:nth-child(2n+1) {
+                    background: transparent !important;
+                }
```
danrot: Why is this change in here at all? Doesn't seem to be connected at all to this PR.

wachterjohannes: no it isnt but the issue happens only in the automation-tab which i have edited also in the linked PR (it happens because of the same reason as i explained here https://github.com/massiveart/husky/pull/756#discussion_r93019251

PR #754, opened 2016-11-28
Added new table-view style

wachterjohannes:
This PR adds a new "light" style for the datagrid table-view.

PR #759, opened 2016-12-16
added inline-block to toggle icon in datagrid

danrot:
This bug causes https://github.com/sulu/sulu/issues/3096.

@wachterjohannes You've removed that while implementing #754, was that by accident or is there a reason for this?


alexander-schranz:
Did test the changes and https://github.com/sulu/sulu/issues/3096 is fixed with this. if it is ok for @wachterjohannes it can be merged.

wachterjohannes:
@danrot no this was introduced by the initial implementation of @nnatter. i have tested it and we can merge it.

PR #760, opened 2016-12-20
Implement basic storage and used it in datagrid

wachterjohannes:
Main PR https://github.com/sulu/sulu/pull/3107


husky_components/datagrid/main.js
```diff
@@ -2175,6 +2184,10 @@
                         url = this.sandbox.uritemplate.expand(uriTemplate, {page: page, limit: limit});
                     }
 
+                    if (this.viewId === 'table') {
```
danrot: Why is that only happening for the  `'table'` view?

husky_components/search/main.js
```diff
@@ -76,6 +76,7 @@ define([], function() {
         initialize: function() {
             this.sandbox.logger.log('initialize', this);
             this.options = this.sandbox.util.extend({}, defaults, this.options);
+            this.storage = storage.get('datagrid', this.options.instanceName);
```
danrot: Shouldn't it say something different than `datagrid`?

husky_services/storage.js
```diff
@@ -0,0 +1,60 @@
...
+    Storage.prototype.getWithDefault = function(key, defaultValue) {
+        if (!this.has(key)) {
+            return defaultValue;
+        }
+
+        return this.get(key);
+    };
+
+    function StorageService() {
+        this.storages = {};
```
danrot: Should that really be accessible via `this`? I would make that "private".

husky_services/storage.js
```diff
@@ -0,0 +1,60 @@
...
+    StorageService.prototype.get = function(type, instanceName) {
+        var key = type + '.' + instanceName;
+        if (!this.storages.hasOwnProperty(key)) {
+            this.storages[key] = new Storage();
+        }
+
+        return this.storages[key];
+    };
+
+    return new StorageService();
```
danrot: Any reason we are doing this the complicated way again? Just returning an object would be easier.

dist/husky.js
```diff
@@ -33403,6 +33463,7 @@ define('husky_components/datagrid/decorators/infinite-scroll-pagination',[],func
                 data: null,
                 instanceName: '',
                 searchInstanceName: null,
+                storageName: null,
```
danrot: Why don't we use the `instanceName` for that? Should also be unique, isn't it?

wachterjohannes: the instanceName is not really unique and it should be possible "enabled" independend from `instanceName`. for example if you use it as a contact list in the organization the `instanceName` could be `contacts` but the `storageName` has to be `contacts-of-account-15` to be unique.

wachterjohannes: and if you dont want to use it in such a list you could deactivate it by leaving `storageName` empty.

husky_components/datagrid/main.js
wachterjohannes: because for other decorator you have to load all the other pages also (e.g. masonry) so i and @chirimoya decided to do it only for table in the first version.

husky_components/search/main.js
wachterjohannes: yes your right should be `'search'` here.

husky_services/storage.js
wachterjohannes: no specific reason i only want to avoid doing it on two different ways in the same file. prototype and returning object. should i change it?

wachterjohannes: i have changed this.

wachterjohannes:
@danrot fixed

dist/husky.js
danrot: In an ideal world the `instanceName` should also be unique, because we open the door for unwanted side-effects otherwise. However, when thinking about it, it would be quite a big BC break, because all the event names change... So I think we leave it as it is.

husky_components/datagrid/main.js
```diff
@@ -2311,6 +2324,11 @@
                             searchFields: this.options.searchFields.join(',')
                         });
                     }
+                    if (searchString !== '') {
+                        this.storage.set('url', url);
+                    } else {
+                        this.storage.remove('url');
```
danrot: Is that the correct behavior? I think this also makes us lose the sorting and that kind of stuff, only because the search field was emptied.

danrot: Or is it not that bad, because a new URL will be generated immediately? That's my impression after testing it.

danrot:
Is this only about remembering the search result? Everything else is already working, right?

husky_components/datagrid/main.js
wachterjohannes: the url will always be saved when the sorting/search is changed therefor all the parameters will be saved.

wachterjohannes:
@danrot not only! the big thing here is to save the url of the datagrid. the search-string saving is only cosmetic to keep the state of the list konsistent with the search-box.

danrot:
There is also a rebase missing.

wachterjohannes:
@danrot done

PR #762, opened 2017-02-02
Fixed datagrid preselect with radio-buttons

wachterjohannes:
Fixes preselect for datagrid with radio-buttons.

Main PR https://github.com/sulu/sulu/pull/3113


scss/modules/form.scss
```diff
@@ -360,7 +360,7 @@ textarea.ckeditor-preview {
             height: $radioSize;
             padding: 2px;
             border-radius: $radioSize;
-            border: 1px solid $borderColor;
+            border: 1px solid $black;
```
danrot: Why are we changing the value here? Why don't we use the more explicit `$borderColor` variable anymore?

wachterjohannes: the border color is medium gray which should be for all the borders except for them which can be clicked (like the radio-button).

PR #763, opened 2017-02-02
fixed empty list for non-table views in datagrid

danrot:
Reverts a small part of the change in https://github.com/massiveart/husky/pull/754. This change caused the empty list coffee icon and the text to be misplaced when used with the masonry view from Sulu.

@wachterjohannes Is there a reason you moved these lines in scss?


wachterjohannes:
tbh not really most of this feature was implemented by @nnatter and i have only taken his commits.

danrot:
@wachterjohannes Can you test and merge it?

PR #769, opened 2017-05-11
added option to concat dropdown url and static items

danrot:
This PR adds an option to the dropdown part of the toolbar component, which allows to change the order in which the items from the URL and the static ones are concatenated.


dist/husky.js
```diff
@@ -38516,7 +38516,8 @@ define('__component__$toolbar@husky',[],function() {
 
                                 // add items if present
                                 if (!!item.dropdownItems) {
-                                    data = data.concat(item.dropdownItems);
+                                    data = !!item.dropdownOptions.concatUrlItemsFirst
```
chirimoya: `concatRequestedItemsFirst`

chirimoya:
else ACK

PR #782, opened 2019-03-04
Fix CKEditor preview for content that is not wrapped in an HTML tag

YetiCGN:
The preview textarea for a property of type `text_editor` in Sulu breaks when the content is not wrapped in an HTML tag, e.g. just plain text without any HTML. This PR changes the code to fallback to the unmodified content in case it cannot be parsed as a DOM node by jQuery.

This occurs for example if a field is switched from `text_area` to `text_editor` and is not converted. 

CKEditor will work just fine and will wrap the content in a new `<p>` tag when editing, but it doesn't get there if the preview errors out before. This change fixes the preview for plain text.


YetiCGN:
I think the npm version used in the tests is too old (v1.2.30) to support the caret version constraints. Current npm version is 6.8.

husky_components/ckeditor/main.js
```diff
@@ -170,15 +170,19 @@ define([], function() {
...
-                $trigger = $(
-                    [
-                        '<textarea class="form-element ckeditor-preview" placeholder="', this.options.placeholder, '">',
-                        text,
-                        '</textarea>'
-                    ].join('')
-                );
+            var text, $trigger;
+            try {
+                text = $(this.$el.val()).text();
```
alexander-schranz: I would do something like `$('<div>' + this.$el.val() + '</div>').text()` because the catch will not work correctly if there is example a tag inside the text.

YetiCGN: I didn't check that case, only theorized about it. I guess using `<body>` instead of `<div>` would be even better. I'll test that.

YetiCGN:
I give up trying to fix the Travis build. This setup is too old to fix in a few minutes. :-D

alexander-schranz:
@YetiCGN the karma build is broken since longer time, and can be ignored.

PR #784, opened 2019-12-05
Add maximum-height to template dropdown

danrot:
fixes sulu/sulu#3061


alexander-schranz:
Behaves now the same way as the language dropdown 👍

